PR #70, opened 2023-08-22
revert: Remove Restyaboard.yml

KieranRobson:
As part of #1
No commits since March 2022 - https://github.com/RestyaPlatform/board
Project also seems to be broken too - https://github.com/RestyaPlatform/board/issues/4426


nodiscc (approved):
Thanks

PR #81, opened 2023-08-24
revert: Remove Volumio.yml

KieranRobson:
As part of #1
Last Updated April 2022 - https://github.com/volumio/Volumio2
Lots of untriaged issues, with some stating that some fuctionality is broken due to volumio 2 being outdated - See   https://github.com/volumio/Volumio2/issues/2192 and https://github.com/volumio/Volumio2/issues/2182


nodiscc (approved):
Thanks

Issue #71, opened 2023-08-22
Filter whole list to show only specific platforms/languages


crispybegs:
as requested by maintainer on reddit.
i'd love a way of being able to filter the whole list to show only those services / apps that have docker containers, since that's how I implement everything

thanks!

nodiscc:
I can see this being useful (e.g someone wanting to manage their entire setup through Docker, or shared hosting users looking for PHP projects...).

The easiest way would be creating a sub-page for each language/platform (similar to how we currently have a subpage for each tag/category). This would need to be implemented in https://github.com/nodiscc/hecat/blob/master/hecat/exporters/markdown_multipage.py

I don't think these sub-pages should be added to the sidebar (cluttering it), however the language tags in the list itself should be made clickable, and link to the corresponding platform/language sub-page.

A special page `All platforms` can also be made available with a link to all platform/language sub-pages.

nodiscc:
Someone at https://old.reddit.com/r/selfhosted/comments/15y7y36/the_awesomeselfhosted_repo_released_a_new_website/ mentioned that it would be useful to have the ability to search for programs that have ready-made Arm binaries available (though [which Arm?](https://nickdesaulniers.github.io/blog/2023/03/10/disambiguating-arm/)).

We could add `Arm` as a platform, and a sub-page would be built, only listing items which match this platform.

KieranRobson:
> We could add Arm as a platform, and a sub-page would be built, only listing items which match this platform.

This seems like the best option, this way we dont need to have different `Arm` platforms and it is more generic for someone who is just browsing for `arm` projects

ShakataGaNai:
Maybe Kubernetes as well? Obviously, if it's dockerized it CAN be run in K8s, but sometimes it can be quite complex if there is no documentation or examples.

nodiscc:
To be clear, this would include any platform listed in https://github.com/awesome-selfhosted/awesome-selfhosted-data/tree/master/platforms, so k8s as well.

cpdeethree:
> Maybe Kubernetes as well? Obviously, if it's dockerized it CAN be run in K8s, but sometimes it can be quite complex if there is no documentation or examples.

Let's be more clear than that, if it's dockerized it can, but if it's an orchestration of containers (e.g. docker-compose) you will need the k8s equivalent which is usually either a Kustomize manifest or a Helm chart (rather than a series of disparate yaml files)

nodiscc:
[...]

PR #117, opened 2023-09-05
platforms: add `deb` platform for software that can be installed trough .deb packages

nodiscc:
- this platform may be added to software that is installable through .deb packages/repositories (either official Debian repositories, or repositories/.deb packages provided by project maintainers)
- related https://github.com/awesome-selfhosted/awesome-selfhosted-data/issues/71
- added the platform to a few items (not exhaustive)

### DIFF
--- a/platforms/deb.yml
+++ b/platforms/deb.yml
@@ -1,0 +1,2 @@
+name: deb
+description: "[deb](https://en.wikipedia.org/wiki/Deb_(file_format)) is the standard package format for Debian GNU/Linux and its derivatives."
--- a/software/asterisk.yml
+++ b/software/asterisk.yml
@@ -5,6 +5,7 @@ licenses:
   - GPL-2.0
 platforms:
   - C
+  - deb
 tags:
   - Communication - SIP
 source_code_url: https://github.com/asterisk/asterisk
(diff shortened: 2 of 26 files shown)


KieranRobson (approved):

kokomo123 (approved):
This does make sense, as it's easier to install deb packages. 👍🏻 💯

Rabenherz112 (approved):
I think it definitely makes sense to list package installations as an installation method. Does it also make sense to consider adding support for other popular package formats, such as RPM (Redhat, Fedora, CentOS, SUSE), if applicable?

kokomo123:
> I think it definitely makes sense to list package installations as an installation method. Does it also make sense to consider adding support for other popular package formats, such as `RPM` (Redhat, Fedora, CentOS, SUSE), if applicable?

I also agree, though that'd be a lot of editing files, I suspect it could be done. I approved this knowing that this could potentially lead to other packaging support, perhaps?

nodiscc:
> Does it also make sense to consider adding support for other popular package formats, such as RPM

I think it does

> though that'd be a lot of editing files

It's a collaborative list so people who are interested in this can put in the required work themselves :) I don't use RH-based distros, but I think RPM is popular enough to accept it as a valid language/platform tag.

PR #124, opened 2023-09-07
Update davical website URL

nodiscc:
- see https://gitlab.com/davical-project/davical/-/issues/301
- ref. #1


nodiscc:
The problem appears to be fixed

Issue #72, opened 2023-08-22
Allow a grace period before removal of unmaintained software that is known to be working


KieranRobson:
One that I found was [`Writing`](https://github.com/josephernest/writing). Flagged here: https://github.com/awesome-selfhosted/awesome-selfhosted-data/actions/runs/5932096580/job/16085261523#step:4:218 

The project seems to still work in the demo test I did. Like this project, there will be many others that are working, and it may be appropriate to keep them within the list/site. 

For these projects, would there be a way to not flag them in the workflow, i.e. add an ignore comment. This may not have been possible in the larger list as it was just one file, however as software is separated into .yml files. I wonder if this would be possible?

nodiscc:
I agree that some projects (unmaintained but simple enough and still working, e.g https://github.com/awesome-selfhosted/awesome-selfhosted/pull/4120) should benefit from a longer "grace period" during which the `awesome-lint` job does not list them in the unmaintained projects report.

This mechanism/"ignore list" should be implemented in https://github.com/nodiscc/hecat/. The question remains, of how we would expect the configuration of such an "ignore"/grace period to look like in the [module configuration](https://github.com/nodiscc/hecat/blob/master/hecat/processors/awesome_lint.py#L9)

I don't think we want to permanently add a project to this ignore list, since it would be good to be reminded to revisit them after a while.  So a simple `ignore_list: ['project1', 'project2', 'writing' ]` might not do the trick.

It may be more appropriate to use a dictionary like

```yaml
grace_periods: # do not issue warning or errors for projects listed here (source_code_url), until the specified date
  - url: https://github.com/josephernest/writing # working despite no updates, https://github.com/awesome-selfhosted/awesome-selfhosted-data/issues/72
    until: 2024-08-22
  - url: https://github.com/another/project
    until: 2023-12-31
```

Other ideas welcome

KieranRobson:
The grace period would be a better idea long term, hence so that they are checked upon and can be extended again if working and removed if not. The dictionary would be a good layout, and easy to contribute to as well. 

I wonder if it is possible to put the dictionary in another file and refer to it, so that the list doesn't grow to be unwieldy.

PR #134, opened 2023-09-08
revert: Remove CUPS.yml

KieranRobson:
As part of #1
No commits since June 2022
Untriaged issues including a couple memory leak ones - https://github.com/apple/cups/issues/6144 and https://github.com/apple/cups/issues/6145


nodiscc:
Given that CUPS has been the standard Unix/Linux printing system since 1999 (https://en.wikipedia.org/wiki/CUPS), is extremely mature and stable, it is expected that it won't receive much maintenance. Additionally, there are no alternatives and it is unlikely there will ever be one. Memory leaks [do not classify](https://owasp.org/www-community/vulnerabilities/Memory_leak) as security issues, but as performance issues; there are currently [no CVE on CUPS packages in Debian](https://security-tracker.debian.org/tracker/source-package/cups).

I'm in favor of keeping this in the list (related https://github.com/awesome-selfhosted/awesome-selfhosted-data/issues/72)

KieranRobson:
Ahh, makes sense! Ill close this and as its linked to the issue mentioned, we can find it again!

PR #137, opened 2023-09-08
Remove Keeweb (unmaintained since June 2022, looking for maintainers)

KieranRobson:
As part of #1
No commits since June 2022 
Multiple issues regarding corruption so probably isnt too stable - https://github.com/keeweb/keeweb/issues/1832 and https://github.com/keeweb/keeweb/issues/2045


nodiscc (approved):
Project looking for maintainers https://github.com/keeweb/keeweb/issues/2022

Thank you

PR #144, opened 2023-09-09
Remove Vanilla-forums (unmaintained since november 2022)

KieranRobson:
As part of #1
No commits since November 2022 however doesnt seem to work in node 16+ - https://github.com/vanilla/vanilla/issues/11180
May not work in 8.2? - https://github.com/vanilla/vanilla/issues/11174


nodiscc (approved):
Thanks

nodiscc:
https://github.com/awesome-selfhosted/awesome-selfhosted-data/commit/b1d8032c93122b1707b539d75444a18762cab242

Blitheness:
I wouldn't call it unmaintained (there have been regular updates this year), but the current company and devs behind it don't seem to be fostering the same level of OSS community support that they once had backing their product.

There is, however, an extra barrier to usability in that they no longer provide build artifacts that can be used to spin up your own forum, you currently have to figure out the correct environment configuration for building their software on your own, so it's probably not in everyone's best interests to continue promoting Vanilla Forums.

Issue #163, opened 2023-09-12
tools/CI: awesome_lint: fail with exit code !=0 when a tag has less than 3 items attached to it


nodiscc:
[Contribution guidelines](https://github.com/awesome-selfhosted/awesome-selfhosted-data/blob/master/CONTRIBUTING.md#add-a-tagcategory) state that `Any tag must have a minimum of 3 software projects referencing it.`. However Forgejo is the only project tagged with `Software Development - Continuous Integration & Deployment`, but it was still added to the resulting page (https://awesome-selfhosted.net/tags/software-development---continuous-integration--deployment.html), defeating the `Please visit awesome-sysadmin/Continuous Integration & Continuous Deployment instead` admonition.

`make awesome_lint` must fail when this occurs.

nodiscc:
In addition, `make awesome_lint` must fail when a project is tagged with a tag for which `redirect` is set (which is the case here https://github.com/awesome-selfhosted/awesome-selfhosted-data/blob/master/tags/software-development---continuous-integration--deployment.yml)

PR #164, opened 2023-09-12
tools/hecat: fail when a software item has a tag for which `redirect` is set

nodiscc:
- https://github.com/nodiscc/hecat/blob/master/hecat/processors/awesome_lint.py#L11
- ref. https://github.com/awesome-selfhosted/awesome-selfhosted-data/issues/163

### DIFF
--- a/.hecat/awesome-lint.yml
+++ b/.hecat/awesome-lint.yml
@@ -4,7 +4,6 @@ steps:
     module: processors/awesome_lint
     module_options:
       source_directory: ./
-      items_in_delegate_to_fatal: False
       licenses_files:
         - licenses.yml
         - licenses-nonfree.yml


nodiscc (commented):
- [x] Usage of `redirect` should be documented in https://github.com/awesome-selfhosted/awesome-selfhosted-data/blob/master/CONTRIBUTING.md#add-a-tagcategory

nodiscc (commented):
Actually, the correct name for this key is `items_in_redirect_fatal`, and is already set to `True` by default. However, it only causes checks to fail when the affected tag is **the first** in the `tags` list.

https://github.com/nodiscc/hecat should be patched so that if a software item uses a tag, that has a non-empty `redirect`, the test should fail regardless of the position of this tag in the list (when `items_in_redirect_fatal: True`)

nodiscc:
Replaced with https://github.com/awesome-selfhosted/awesome-selfhosted-data/pull/177

PR #167, opened 2023-09-17
Corrections to Rallly data

lukevella:
Hi, I've updated the information for Rallly as some of it was a bit out of date and it is currently listed in the wrong category.


software/rallly.yml
```diff
@@ -1,13 +1,16 @@
 name: Rallly
 website_url: https://rallly.co
-description: Create polls to vote on dates and times (alternative to Doodle).
+description: Open-source Doodle alternative. Poll for availability and schedule group meetings.
```
KieranRobson, line 3: Please keep the (alternative to Doodle) as this is standard throughout most of the descriptions

KieranRobson (changes requested):
Quick change to be made

software/rallly.yml
lukevella: Done 👍

nodiscc (changes requested):
I think the previous description was actually better - if you really think it should be changed, why not use the official description from https://github.com/lukevella/rallly/blob/main/README.md?

> Schedule group meetings with friends, colleagues and teams. Create meeting polls to find the best date and time to organize an event based on your participants' availability.

lukevella:
Sure, I've reverted the change to the description then 👍

software/rallly.yml
```diff
@@ -4,10 +4,13 @@ description: Create polls to vote on dates and times (alternative to Doodle).
 licenses:
   - AGPL-3.0
 platforms:
-  - Nodejs
+  - Node.js
```
nodiscc, line 7:
```suggestion
  - Nodejs
```

The previous syntax was correct https://github.com/awesome-selfhosted/awesome-selfhosted-data/blob/master/platforms/nodejs.yml

nodiscc (changes requested):

nodiscc (approved):
Thanks

PR #170, opened 2023-09-18
Remove Pomf (Unmaintained since April 2022)

KieranRobson:
As part of #1
No code  commits since April 2022
Maintainer recommends using a different project of theirs -https://github.com/pomf/pomf/commit/c6806f4db6858d4331483d328088cae56281f472


nodiscc (approved):
Thanks

PR #182, opened 2023-09-21
Remove Feedka (Unmaintained since November 2022, Outdated dependencies)

KieranRobson:
As part of #1
No commits since November 2022
Recommended ruby and rails versions hit EOL - https://drabkirn.cdadityang.xyz/docs#feedka - https://endoflife.date/ruby
Low star count suggests not a ton of uses to report issues that may come from eol docs


nodiscc (approved):
thanks

PR #202, opened 2023-09-23
Fix Guacamole source code

Rabenherz112:
- Guacamole was donated to the [Apache Software Foundation](http://apache.org/) in 2016 and is now [Apache Guacamole}(https://guacamole.apache.org/)
- Repo is archived


nodiscc (approved):
Thanks

PR #206, opened 2023-09-23
Remove pootle (using unsupported python version)

Rabenherz112:
- Requires Python2 to run which is not supported anymore ([Sunsetting Python 2](https://www.python.org/doc/sunset-python-2/))

ref: #1


nodiscc (approved):
thanks

PR #228, opened 2023-10-03
Remove Livekit

nodiscc:
- Livekit is not an end-user application, it is a WebRTC development toolkit meant to be integrated in video conferencing applications.
- As per multiple previous issues/PRs, we only list end-user applications, not SDKs or libraries


kokomo123 (approved):
No issues with this as yes, we need to slim down the list to include only end-user tools.

PR #234, opened 2023-10-03
Add Manufacturing tag

hasecilu:
New category to allocate Octoprint & more.

Thanks for taking the time to suggest an addition to awesome-selfhosted! 

To ensure your Pull Request is dealt with swiftly, please check the following (check the boxes `[x]`):
- [X] Submit one item per pull request. This eases reviewing and speeds up inclusion.
- [X] You have searched the repository for any relevant [issues](https://github.com/awesome-selfhosted/awesome-selfhosted-data/issues) or [PRs](https://github.com/awesome-selfhosted/awesome-selfhosted-data/pulls), including closed ones.
- [X] Any software you are adding is not already listed at any of [awesome-sysadmin](https://github.com/n1trux/awesome-sysadmin), [staticgen.com](https://www.staticgen.com/), [staticsitegenerators.net](https://staticsitegenerators.net/), [dbdb.io](https://dbdb.io/browse).
- [X] The file you are adding is formatted as described in [addition.md](https://github.com/awesome-selfhosted/awesome-selfhosted-data/blob/master/.github/ISSUE_TEMPLATE/addition.md).
- [X] `Demo` links should only be used for interactive demos, i.e. not video demonstrations. 
- [X] Comments and unused optional fields have been removed.
- [X] The file you are adding uses [kebab-case](https://en.wikipedia.org/wiki/Letter_case#Kebab_case) file naming, for example `my-awesome-software.yml`.
- [X] Values for `platform` are the main **server-side** requirements for the software. Don't include frameworks or specific dialects.
- [X] Any software project you are adding to the list is actively maintained.
- [X] Any software project you are adding was first released more than 4 months ago.
- [X] Any software project you are adding has working installation instructions.
- [X] If you are adding software forked from another active project, please provide/link to a clear list of differences between both.
- [X] You understand that your Pull Request will be merged ~1 week after approval, to allow for further comments if needed.

PR #239, opened 2023-10-04
Move VanDam to Manufacturing

KieranRobson:
Depends on #234
Could not comment on lines so created this 

Ruby is only used in dev so removed from platforms. Cleaned up description so it flowed better.


nodiscc (approved):

nodiscc:
Rebased, squashed and merged in https://github.com/awesome-selfhosted/awesome-selfhosted-data/pull/268. Thanks

PR #247, opened 2023-10-09
Change DailyTxt tag to Note-taking & Editors

nodiscc:
> DailyTxT is an encrypted Diary Web-App to write down your stories of the day and to find them again easily.


KieranRobson (approved):
> Encrypted Diary Web-App

Good catch! Description could be made to be sentence case however this isn't a major change needed.

PR #256, opened 2023-10-09
Remove Shinobi (Unmaintained since Feb 2021)

KieranRobson:
As part of #1
Unmaintained since Feb 2021
Mirror also showing no commits 

https://github.com/ShinobiCCTV/Shinobi


PrplHaz4:
Shinobi moved to gitlab a while ago, with the last commit to master being 1 month ago, dev 1 week ago:
https://gitlab.com/Shinobi-Systems/Shinobi

KieranRobson:
> Shinobi moved to gitlab a while ago, with the last commit to master being 1 month ago, dev 1 week ago: [gitlab.com/Shinobi-Systems/Shinobi](https://gitlab.com/Shinobi-Systems/Shinobi)

Good catch, the CE (https://gitlab.com/Shinobi-Systems/ShinobiCE) was on GitLab also however seemed to be abandoned in favour for the the above repo that is not free (?) https://gitlab.com/Shinobi-Systems/Shinobi/-/blob/master/COPYING.md

nodiscc (approved):
The CE repository appears to be a GPLv3/AGPLv3-licensed version of the non-CE repository, which uses a custom license that prohibits commercial use (hence non-[Free](https://en.wikipedia.org/wiki/Free_software#Definition_and_the_Four_Essential_Freedoms_of_Free_Software)). I think removing the CE repository from awesome-selfhosted is the correct thing to do, given that it's unmaintained. The non-Free edition can be added later.

PR #258, opened 2023-10-09
Update Dovel Source Code Link

KieranRobson:
As part of #1 
There doesnt seem to be any of the project files (?) - https://git.derelict.garden/dovel/email/tree


nodiscc (changes requested):
`source_code_url` should be changed to https://dovel.email/server/tree.html as it appears the repository was moved (found the new address at https://dovel.email/git.html).

nodiscc (approved):
Thanks again

PR #260, opened 2023-10-10
Update URL for Tiki source code

marclaporte:
Thank you SourceForge for all the support over the years.

https://dev.tiki.org/Tiki-Version-control-history


nodiscc (approved):
Thanks

PR #265, opened 2023-10-10
Remove Bluemind (Dead Source Code)

KieranRobson:
Happy to change to None-free rather than removal however

The source code link is no longer active, no alternative found here: https://doc.bluemind.net/release/4.9/en/guide_du_developpeur/construction_de_bluemind

To download it, you have to submit details such as email and postal code. - https://download.bluemind.net/bm-download/register


nodiscc (approved):
Thanks

PR #266, opened 2023-10-10
Remove OpenGTS (Unmaintained since Feb 2020)

KieranRobson:
As part of #1
Unmaintained since Feb 2022

Source code leads to here: https://sourceforge.net/projects/opengts/files/server-base/ where 2.6.5 was released in 2017 however http://www.opengts.org/ shows a 2020 release. 

http://track.opengts.org/track/Track Demo requires login with no details I could find


nodiscc (approved):
Thanks

PR #275, opened 2023-10-12
add Video Surveillance tag

nodiscc:
- move 4 items from Video Streaming to Video Surveillance
- improve related tags


KieranRobson (approved):
I like it!

PR #278, opened 2023-10-12
Remove Calypso (Unmaintained since May 2021)

KieranRobson:
As part of #1
Unmaintained since May 2021 - https://keithp.com/cgit/calypso.git/


nodiscc (approved):
Thanks

PR #294, opened 2023-10-18
Remove Calagator (Unmaintained since Dec 2022)

KieranRobson:
As part of #1

Doesnt seem to be much activity see: https://github.com/calagator/calagator/pull/721#issuecomment-1500743904 and https://github.com/calagator/calagator/issues/705

Requires Rails 5 which is EOL along woth Ruby 2.6


kokomo123 (approved):

nodiscc (approved):
Thanks

nodiscc:
460eff451

PR #302, opened 2023-10-21
Remove Silverstrike (Unmaintained, Dead Links)

KieranRobson:
As part of  #1
No commuts since Dec 2022
https://silverstrike.org - Timesout 
https://demo.silverstrike.org - Timesout


nodiscc (approved):
Thanks

PR #311, opened 2023-10-23
Remove Euterpe (Outdated ICU Version)

KieranRobson:
As part of #1
Outdated ICU Version from April 2022 - https://github.com/ironsmile/euterpe/issues/54 / https://icu.unicode.org/download/71#TOC-ICU4C-Download


nodiscc (approved):
Thanks

PR #313, opened 2023-10-23
Remove Eonza Demo

KieranRobson:
Demo redirected to this link

### DIFF
--- a/software/eonza.yml
+++ b/software/eonza.yml
@@ -8,7 +8,7 @@ platforms:
 tags:
   - Automation
 source_code_url: https://github.com/gentee/eonza
-demo_url: https://playground.eonza.org/
+demo_url: https://playground.eonza.org:4010/#/
 stargazers_count: 124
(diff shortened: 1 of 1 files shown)


kokomo123 (approved):

nodiscc (changes requested):
See #312, I don't think URLs should be updated based on redirects, as long as the domain name stays the same.

In this particular case I get an error when visiting the demo URL:

```
$ curl https://playground.eonza.org/
curl: (60) SSL certificate problem: certificate has expired
More details here: https://curl.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
```

In this case I think it is better to remove the demo link completely.

kokomo123:
fair fair, will know this for next time

nodiscc (approved):
Thanks

PR #317, opened 2023-10-26
Remove Thredded (Outdated Ruby Version)

KieranRobson:
As part of #1
Outdated ruby version - https://github.com/thredded/thredded/issues/978

Outdated docker - https://github.com/thredded/thredded/issues/977


nodiscc (approved):
Thanks

PR #319, opened 2023-10-27
Remove DocPHT (Unmaintained, Outdated PHP Version)

KieranRobson:
As past of #1
Unmaintained since 2020
Php version 7.2


nodiscc (approved):
Thanks.

PR #242, opened 2023-10-06
Delete inactive homepage by tomershvueli

nimodius:
Project seems outdated and confusing with other active one with name that is same

Thanks for taking the time to suggest an addition to awesome-selfhosted! 

To ensure your Pull Request is dealt with swiftly, please check the following (check the boxes `[x]`):
- [x] Submit one item per pull request. This eases reviewing and speeds up inclusion.
- [x] You have searched the repository for any relevant [issues](https://github.com/awesome-selfhosted/awesome-selfhosted-data/issues) or [PRs](https://github.com/awesome-selfhosted/awesome-selfhosted-data/pulls), including closed ones.

Issue #1, opened 2021-06-27
Remove dead links, unmaintained projects


nodiscc:

king861:
Gave me you WhatsApp number

mozher:
1800 : AT!BAND=03,"1800",0,4
2100 : AT!BAND=04,"2100",0,1
700  : AT!BAND=05,"700",0,8000000
1800 + 700: AT!BAND=06,"1800 +700",0,8000004
1800 + 2100: AT!BAND=07,"1800 +2100",0,5
2100 + 700 : AT!BAND=08,"2100 +700",0,8000001
1800 + 2100 + 700: AT!BAND=09,"1800 +2100 +700",0,8000005
1800 + 2100 + 2300: 8000000005
1800 + 2300: 8000000004
2300: 8000000000
2100 + 2300: 8000000001
700 + 2300: 8008000000
700 + 2100 + 2300: 8008000001
700 + 1800 + 2300: 8008000004

اكواد ترددات Mobily:
800 : 80000
1800 : AT!BAND=03,"1800",0,4
2100 : AT!BAND=04,"2100",0,1
2500 : AT!BAND=05,"2500",0,10000000000
2600 : AT!BAND=06,"2600",0,2000000000
1800 + 2100: AT!BAND=07,"1800 +2100",0,5
2500 + 2600: AT!BAND=08,"2500 +2600",0,12000000000
1800 + 2100 + 800: AT!BAND=09,"1800 +2100 +800",0,80005
1800 + 2100 + 2500: 10000000005
2100 + 2600 : 2000000001
1800 + 2500 : 10000000004
1800 + 2600 : 2000000004
2100 + 2500 : 10000000001
1800 + 2100 + 2600 : 2000000005
1800 + 2500 + 2600 : 12000000004
2100 + 2500 + 2600 : 12000000001
Zain:
1800 : AT!BAND=03,"1800",0,4
2100 : AT!BAND=04,"2100",0,1
900  : AT!BAND=05,"900",0,80
1800 + 900: AT!BAND=06,"1800 +900",0,84
1800 + 2100: AT!BAND=07,"1800 +2100",0,5
2100 + 900: AT!BAND=08,"2100 +900",0,81
1800 + 2100 + 900: AT!BAND=09,"1800 +2100 vgv+900",0,85
800 : 80000
2600 : 2000000000

powru9-gobVok-nutguj:
[]()

[...]

PR #322, opened 2023-10-27
tools/hecat: do not warn about last update date for homepage, posio, bicimon, cubiks-2048

nodiscc:
- these projects are simple enough and have no external dependencies, so no maintenance is expected
- ref. https://github.com/awesome-selfhosted/awesome-selfhosted-data/issues/72
- ref. https://github.com/awesome-selfhosted/awesome-selfhosted-data/pull/242

### DIFF
--- a/.hecat/awesome-lint-strict.yml
+++ b/.hecat/awesome-lint-strict.yml
@@ -10,3 +10,8 @@ steps:
       licenses_files:
         - licenses.yml
         - licenses-nonfree.yml
+      last_updated_skip: # (optional, default []) list of items (source_code_url) for which the last update date check should not produce errors/warnings
+        - https://github.com/tomershvueli/homepage # simple/no maintenance required https://github.com/awesome-selfhosted/awesome-selfhosted-data/pull/242
+        - https://github.com/abrenaut/posio # simple/no maintenance required
+        - https://github.com/knrdl/bicimon # simple/no maintenance required
+        - https://github.com/Kshitij-Banerjee/Cubiks-2048 # simple/no maintenance required
--- a/.hecat/awesome-lint.yml
+++ b/.hecat/awesome-lint.yml
@@ -7,3 +7,8 @@ steps:
       licenses_files:
         - licenses.yml
         - licenses-nonfree.yml
+      last_updated_skip: # (optional, default []) list of items (source_code_url) for which the last update date check should not produce errors/warnings
+        - https://github.com/tomershvueli/homepage # simple/no maintenance required https://github.com/awesome-selfhosted/awesome-selfhosted-data/pull/242
+        - https://github.com/abrenaut/posio # simple/no maintenance required
+        - https://github.com/knrdl/bicimon # simple/no maintenance required
+        - https://github.com/Kshitij-Banerjee/Cubiks-2048 # simple/no maintenance required


nodiscc:
As expected, no warning/error is raised for projects listed in `last_updated_skip`:

```
INFO:awesome_lint.py: Bicimon: skipping last update time check as per configuration (last_updated_skip) (-561 days, 7:11:27.185372)
INFO:awesome_lint.py: Cubiks-2048: skipping last update time check as per configuration (last_updated_skip) (-3430 days, 7:11:27.172919)
INFO:awesome_lint.py: Homepage by tomershvueli: skipping last update time check as per configuration (last_updated_skip) (-902 days, 7:11:27.152919)
INFO:awesome_lint.py: Posio: skipping last update time check as per configuration (last_updated_skip) (-744 days, 7:11:27.116016)
```

Note that the exclusions must be added to both `.hecat/awesome-lint.yml` (which runs for pull requests and commits on `master` but does not block the build) and `.hecat/awesome-lint-strict.yml` (which runs nightly and produces the badge  which indicates whether the list currently contains unmaintained projects.

kokomo123 (approved):
As long as they are simple in question, and they don't have any significant vulnerabilities (or will in the future), it looks fine to me!

KieranRobson (approved):
Nice and easy setup to contribute to!

Rabenherz112 (approved):
I largely agree with your selections and I'm pleased to see that we now have a means to exclude projects from the scan. However, I recall that your initial plan was not to permit permanent additions to a "whitelist."

> I don't think we want to permanently add a project to this ignore list, since it would be good to be reminded to revisit them after a while. So a simple ignore_list: ['project1', 'project2', 'writing' ] might not do the trick.

([Source](https://github.com/awesome-selfhosted/awesome-selfhosted-data/issues/72#issuecomment-1688461538))

Is there any reason for this change, as I personally believe your idea of incorporating the 'until: 2024-08-22' feature was quite promising.

KieranRobson:
As they are still in the list and the time is visible, Id say the current set up is fine.

nodiscc:
>  I recall that your initial plan was not to permit permanent additions to a "whitelist." [...] your idea of incorporating the 'until: 2024-08-22' feature was quite promising.

I opted for a simpler implementation in the first time, which is also simpler to use. My concern with permanently adding projects to the ignore list, was that we might forget to check them again after some time. But the following points lead me to think it will not be a problem:
- The 4 projects I just added are static HTML/JS applications with no external dependencies, no server-side code, and will basically never require any kind of maintenance.
- Projects that match the ignore list are logged in hecat/github actions output, so they are not _completely_ silently ignored. I will probably check the logs linked from https://github.com/awesome-selfhosted/awesome-selfhosted-data/issues/1 and review these at least once a year, which is sufficient in my opinion. These projects' last update dates are still rendered in red on https://awesome-selfhosted.net/ which gives an additional hint.
- In cases where a project is actually in a _temporary_ grace period (e.g https://github.com/awesome-selfhosted/awesome-selfhosted/pull/4120), we can keep an issue or pull request open and tag it `on hold` (this particular one was closed during the migration to `awesome-selfhosted-data` but I will make a note to recheck on 21 November).

PR #330, opened 2023-11-01
Add Evergreen to Lint Ignore List

KieranRobson:
```
        - https://github.com/evergreen-library-system/Evergreen # Not actually unmaintained
```

Both have commits in the last week on their master branch.

### DIFF
--- a/.hecat/awesome-lint-strict.yml
+++ b/.hecat/awesome-lint-strict.yml
@@ -15,6 +15,5 @@ steps:
         - https://github.com/abrenaut/posio # simple/no maintenance required
         - https://github.com/knrdl/bicimon # simple/no maintenance required
         - https://github.com/Kshitij-Banerjee/Cubiks-2048 # simple/no maintenance required
-        - https://github.com/NodeBB/NodeBB # Not actually unmaintained 
         - https://github.com/evergreen-library-system/Evergreen # Not actually unmaintained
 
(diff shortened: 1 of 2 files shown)


kokomo123 (approved):
sounds good as long as it's being committed to! LGTM.

nodiscc:
The latest check [[1]](https://github.com/awesome-selfhosted/awesome-selfhosted-data/actions/runs/6725820472/job/18280936713) reports `ERROR:awesome_lint.py: Evergreen: last updated -421 days, 1:31:07.683128 ago, older than 365 days`, and [`evergreen.yml`](https://github.com/awesome-selfhosted/awesome-selfhosted-data/blob/master/software/evergreen.yml) has `updated_at: '2022-09-07'` but this is incorrect: the `main` branch which is the default branch was updated 2 days ago: https://github.com/evergreen-library-system/Evergreen/commits/main.

I think there is a bug in the automatic metadata update process (hhttps://github.com/awesome-selfhosted/awesome-selfhosted-data/commit/68b1d8200d8cb9d2319a13ece092eaecbf83976c#diff-631f755b6dcfcff38966bbc177c58041f400d35b3f227c0d633733071ef591b6)

In this case, an issue should be opened about this bug and we should not push it under the rug by just adding it to the ignore list.

nodiscc:
Moved to https://github.com/awesome-selfhosted/awesome-selfhosted-data/issues/332, thanks for noticing.

PR #334, opened 2023-11-03
Added links to URL Check and Removed Psychic  URL

KieranRobson:
Added Moodle and Open Web Analytics
Rmoved Psychic Dev URL

All links have been showing isses for the past week ish. It looks like moodle is not a temporary issue as hoped!


nodiscc (approved):
Thanks, indeed it looks like these links always return 403 when queried from github actions

PR #356, opened 2023-11-10
Remove depends_3rdparty for Flyimg project

sadok-f:
Remove depends_3rdparty for Flyimg project to remove the flag ⚠ Anti-features

### DIFF
--- a/software/flyimg.yml
+++ b/software/flyimg.yml
@@ -8,7 +8,6 @@ platforms:
   - Docker
 tags:
   - Miscellaneous
-depends_3rdparty: no
 demo_url: https://demo.flyimg.io
 stargazers_count: 834
 updated_at: '2023-10-30'


kokomo123:
Uhh, you didn't actually remove the depends_3rdparty line for flyimg project, and can you give a description as to why you are removing it's dependency as it doesn't depend on 3rd partys, or..?

nodiscc:
Hi, your commit is empty: https://github.com/awesome-selfhosted/awesome-selfhosted-data/pull/356/commits/c6d391bc45857f3939e7f55f6d3d3ae407645e6d

There is something strange going on: https://awesome-selfhosted.net/#flyimg is marked with `⚠ Anti features`, however https://github.com/awesome-selfhosted/awesome-selfhosted-data/blob/master/software/flyimg.yml#L11 states `depends_3rdparty: no`, so the Anti-features label shouldn't be there in the first.

I'll investigate

sadok-f:
Sorry, I'm updating my PR to remove the depends_3rdparty option, it seems when setting the option to `no` it shows the flag ⚠ Anti-features

nodiscc:
> it seems when setting the option to no it shows the flag ⚠ Anti-features

Yes it seems like so. This is a bug, I'll look into it.

Keeping this open since removing `depends_3rdparty: no` is also correct (it's the default value, at least in theory). Thanks

nodiscc:
Fixed in https://github.com/awesome-selfhosted/awesome-selfhosted-data/pull/365

https://awesome-selfhosted.net/#flyimg

Thanks again

sadok-f:
Great, thank you 👍

PR #383, opened 2023-11-18
Remove Pico (unmaintained since July 2022)

nodiscc:
- ref. #1 
- `ERROR:awesome_lint.py: Pico: last updated -479 days, 1:29:43.500619 ago, older than 365 days`
 compatibility problems https://github.com/picocms/Pico/issues/608
- Unaddressed issues about outdated/vulnerable dependencies https://github.com/picocms/Pico/issues/679
- broken CI


nodiscc:
e895360e9

PR #402, opened 2023-11-26
Update Budibase License

nodiscc:
- Budibase cannot be built from GPL-3.0 sources https://github.com/Budibase/budibase/discussions/12377, https://github.com/Budibase/budibase/blob/master/LICENSE
- The next version will enforce an artificial limit on the number of users, which cannot be patched out https://budibase.com/blog/updates/pricing-v3/


Ki-er (approved):

kokomo123 (approved):
LGTM, no issues here. Thanks for updating the license!

PR #405, opened 2023-11-28
Remove Django Todo (Unmaintained since November 2022)

Ki-er:
As part of #1

`ERROR:awesome_lint.py: django-todo: last updated -390 days, 1:31:10.952409 ago, older than 365 days`

Issues with installing leads to convoluted work around - https://github.com/shacker/django-todo/issues/140#issuecomment-1817511134


nodiscc:
8af5a9b4a

PR #408, opened 2023-12-01
Update DocuSeal

AlexandrToorchyn:
Improve DocuSeal app description

### DIFF
--- a/software/docuseal.yml
+++ b/software/docuseal.yml
@@ -1,10 +1,11 @@
-name: Docuseal
+name: DocuSeal
 website_url: https://www.docuseal.co
 source_code_url: https://github.com/docusealco/docuseal
-description: Create, fill, and sign digital documents (alternative to DocuSign).
+description: Self-hosted DocuSign alternative. Create, fill, and sign digital documents.
 licenses:
   - AGPL-3.0
 platforms:
+  - Ruby
   - Docker
 tags:
   - Document Management


Ki-er:
The alternative to docusign is a standard across all projects.

AlexandrToorchyn:
Hm.. it doesn't look like the list of "Document Management" projects contains DocuSign alternatives. There many projects for storing/manging/manipulating documents, but only DocuSeal in this category is a DocuSign alternative for document eSigning.
Does this address your question?

Ki-er:
Sorry for the confusion, What I meant is that the (alternative to x) is standard.

Ki-er:
Closing as this changes a standard throughout this repo

nodiscc:
@Ki-er is correct, `(alternative to $PROPRIETARY_SERVICE)` is the standard syntax for all projects that present themselves as alternatives to a SaaS or closed-source product.

This is not documented anywhere though, so I'm starting to think we need a document for this kind of "rarely" used guidelines:
- `If you are adding software forked from another active project, please provide/link to a clear list of differences between both.` (move this out of the pull request template to make it easier to read) 
- `If the the project has no documentation in English, please add (documentation in LANGUAGE) at the end of the description.` (currently undocumented)
- `If the project is presented as an alternative to another service or application, please mention it as (alternative to PRODUCT1, RPDUCT2) at the end of the description` (undocumented)
- There could be some sort of guideline about how to phrase descriptions too (for example `Minimalist text adventure game` would be preferred to `A minimalist text adventure game` or `$PROJECT is a minimalist text adventure game`)

Just leaving this here until we find a way to integrate this.

PR #488, opened 2024-01-10
Move changedetection.io to Automation

nodiscc:
This software is a good match for the Automation section (_Automation software designed to reduce human intervention in processes._) - moving it there also reduces the number of projects in the _Miscellaneous_ (a.k.a. unsorted for lack of an appropriate category) category.


Rabenherz112 (approved):
Makes a lot of sense!

PR #501, opened 2024-01-16
Update janus demo url

Rabenherz112:
- ref: #1 
- `https://janus.conf.meetecho.com/demos.html : HTTP 404`
- https://github.com/meetecho/janus-gateway/pull/3302


nodiscc (approved):
Thanks

PR #515, opened 2024-01-20
Fix easyappointments demo uri

Rabenherz112:
- `https://easyappointments.org/demo/ : HTTP 404`
- Demo got moved to an own subdomain


nodiscc (approved):
Thanks

PR #524, opened 2024-01-23
Update mattermost url

Rabenherz112:
- ref: #1
- They moved their tld from .org to .com
- A redirect is currently still set-up


nodiscc (approved):
Thanks

PR #550, opened 2024-01-30
Adding docker tag for anonaddy

ryangurn:
"For those who prefer using Docker there is an image you can use here - github.com/anonaddy/docker."

They provide an official docker image and link to it from their repository.


Ki-er (approved):
Thanks!

nodiscc (approved):
Thanks

PR #591, opened 2024-02-28
Add end user specificity

Ki-er:
Enhance the description for platforms to specify end user requirement. 

I wanted to add the instruction requirement however the checklist is already unwieldy and I didnt want to make it even more so. 

Could put this in the other instructions alongside https://github.com/awesome-selfhosted/awesome-selfhosted-data/blob/master/CONTRIBUTING.md#other-guidelines.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -11,7 +11,7 @@ To ensure your Pull Request is dealt with swiftly, please check the following (c
 - [ ] `Demo` links should only be used for interactive demos, i.e. not video demonstrations. 
 - [ ] Comments and unused optional fields have been removed.
 - [ ] The file you are adding uses [kebab-case](https://en.wikipedia.org/wiki/Letter_case#Kebab_case) file naming, for example `my-awesome-software.yml`.
-- [ ] Values for `platform` are the main **server-side** requirements for the software. Don't include frameworks or specific dialects.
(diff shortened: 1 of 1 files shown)


kokomo123:
Other guidelines seems like the most proper place considering the last PR we did to not make this list more daunting.

nodiscc:
In my opinion the guideline cannot be made clearer

> Values for `platform` are the main **server-side** requirements for the software

Setting a non-existing platform will cause an error in automated checks, so the submitter will receive the warning and hopefully investigate. We could make the automated check results more helpful, when this error is raised, by printing out a list of valid platforms (from https://github.com/awesome-selfhosted/awesome-selfhosted-data/tree/master/platforms)

There are a few exceptions to this rule, where someone can submit a new entry with platforms that are valid/in the list, but actually incorrect:
- HTML/Javascript: There are no server-side requirements, except the ability to serve webpages
- Languages that compile to static binaries (C/Go/Rust/...) that just require a libc, an OS and a CPU. I think it is appropriate to still list them with their programming language as platform, instead of having something ambiguous such as `None`, `amd64`, etc.
- Automation tools/scripts that deploy multiple components (shell scripts, ansible), for which it is clearer to use the tool name as a platform

We could detail this in https://github.com/awesome-selfhosted/awesome-selfhosted-data/blob/master/CONTRIBUTING.md#other-guidelines as @kokomo123 suggested, because 1. it's too long to explain in the PR template 2. submitters make these mistakes in good faith, so in the end, someone still has to review the PR - I don't think this change will help

Ki-er:
I agree with all the above points and therefore can just close this PR. Point 2 especially, we are reviewing anyway and can pick up these mistakes!

PR #592, opened 2024-02-29
Update kimai license

richin13:
**Description**

Updating the license since it now uses AGPL v3 https://github.com/kimai/kimai/blob/main/LICENSE

### DIFF
--- a/software/kimai.yml
+++ b/software/kimai.yml
@@ -2,7 +2,7 @@ name: Kimai
 website_url: https://www.kimai.org/
 description: Kimai is a free & open source timetracker. It tracks work time and prints out a summary of your activities on demand.
 licenses:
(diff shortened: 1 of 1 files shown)


software/kimai.yml
```diff
@@ -2,7 +2,7 @@ name: Kimai
 website_url: https://www.kimai.org/
 description: Kimai is a free & open source timetracker. It tracks work time and prints out a summary of your activities on demand.
 licenses:
-  - MIT
+  - AGPL v3
```
nodiscc, line 5:
```suggestion
  - AGPL-3.0
```

nodiscc (changes requested):

nodiscc (approved):
thanks

Issue #568, opened 2024-02-15
Add Jauth


Jipok:
Thanks for taking the time to suggest an addition to awesome-selfhosted! 

Please fill out information below (all fields are mandatory unless noted otherwise):

```yaml
name: "Jauth"
website_url: "https://github.com/Jipok/Jauth"
source_code_url: "[https://gitlab.com/awesome/software"](https://github.com/Jipok/Jauth)
description: "Lightweight SSL/TLS reverse proxy with authorization(via Telegram and SSH) for self-hosted apps."
licenses:
  - GPL-3.0
platforms:
  - Go
tags:
  - Proxy
demo_url: "https://j.ateam.undo.it/"
```

To ensure your issue is dealt with swiftly, please check the following (check the boxes `[x]`):
- [x] Submit one item per pull issue. This eases reviewing and speeds up inclusion.
- [x] You have searched the repository for any relevant [issues](https://github.com/awesome-selfhosted/awesome-selfhosted-data/issues) or [PRs](https://github.com/awesome-selfhosted/awesome-selfhosted-data/pulls), including closed ones.
- [x] Any software you are adding is not already listed at any of [awesome-sysadmin](https://github.com/n1trux/awesome-sysadmin), [staticgen.com](https://www.staticgen.com/), [staticsitegenerators.net](https://staticsitegenerators.net/), [dbdb.io](https://dbdb.io/browse).
- [x] Any software project you are adding to the list is actively maintained.
- [x] Any software project you are adding was first released more than 4 months ago.
- [x] Any software project you are adding has working installation instructions.
- [ ] If you are adding software forked from another active project, please provide/link to a clear list of differences between both.

Jipok:
@nodiscc 
Sorry for ping. But why Anti-features?
<details><summary>screenshot</summary>

![image](https://github.com/awesome-selfhosted/awesome-selfhosted-data/assets/25588359/142a6eef-cbd1-4fbb-98f0-0f00c0effc25)

</details>

nodiscc:
@Jipok hover your mouse cursor on the label, look at the tooltip

> Depends on a proprietary service outside the user's control

Dependency on Telegram

Jipok:
> Dependency on Telegram
Oh. This is an optional dependency. But thanks for the clarification.
Now in retrospect this seems obvious.

PR #595, opened 2024-03-01
Remove Demo URL for JAuth

ryangurn:
https://j.ateam.undo.it/ is responding with a 401 which is causing the deadlinks action to error, additionally the demo url is not listed anywhere in github for jauth. 

For that reason I propose removing the demo url.

Referencing #568, #570.

### DIFF
--- a/software/jauth.yml
+++ b/software/jauth.yml
@@ -8,7 +8,6 @@ platforms:
   - Go
 tags:
   - Proxy
-demo_url: https://j.ateam.undo.it/
 depends_3rdparty: true
 stargazers_count: 11
 updated_at: '2024-02-24'


nodiscc (approved):
Thank you

kokomo123:
Sorry to bump a closed MR--it does redirect supposedly to this site: https://ateam.undo.it/jauth-sso/j.ateam.undo.it/

Jipok:
I can add a link to the demo site in the readme. I could also make a special demo version of the application that allows everyone to log in. But the server will still respond with `401 Unauthorized` since this makes a lot of sense for mobile clients (and not only).

kokomo123:
Feel free to move this to a PR or Issue @Jipok, I was trying to explain I only got it from your initial issue and it works fine there, but yeah, making it visible and special demo will make sense. We can also add it to the exemption list perhaps.

PR #602, opened 2024-03-06
Fix Daneecloud (404. Login required for demo)

Ki-er:
Fix Daneecloud main site - Old showed 404
Remove demo as I couldn't find any login details.


nodiscc (approved):
thanks

PR #643, opened 2024-04-02
Fix pretix metadata

Techwolf12:
Fix the repository, add extra platform/tag info

### DIFF
--- a/software/pretix.yml
+++ b/software/pretix.yml
@@ -6,7 +6,6 @@ licenses:
 platforms:
   - Python
   - Docker
-  - K8S
 tags:
   - E-commerce
   - Polls and Events
--- a/software/pretix.yml
+++ b/software/pretix.yml
@@ -5,6 +5,9 @@ licenses:
   - Apache-2.0
 platforms:
   - Python
+  - Docker
+  - K8S
 tags:
   - E-commerce
-source_code_url: https://github.com/pretix
+  - Polls and Events
+source_code_url: https://github.com/pretix/pretix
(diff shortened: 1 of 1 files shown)


software/pretix.yml
```diff
@@ -5,6 +5,9 @@ licenses:
   - Apache-2.0
 platforms:
   - Python
+  - Docker
+  - K8S
```
nodiscc, line 9:
```suggestion
  - K8S
```

I see no mention of K8S in the documentation

nodiscc (changes requested):

software/pretix.yml
Techwolf12: Still pending a merge: https://github.com/pretix/pretix/pull/4011
But here is the helm chart artifacthub: https://artifacthub.io/packages/helm/techwolf12/pretix

nodiscc: Sure, but it can only be added when `doc/admin/installation/kubernetes.rst` is merged into the upstream documentation, please remove this platform for the time being. Thanks

software/pretix.yml
```diff
@@ -5,6 +5,8 @@ licenses:
   - Apache-2.0
 platforms:
   - Python
+  - Docker
 tags:
   - E-commerce
-source_code_url: https://github.com/pretix
+  - Polls and Events
```
nodiscc, line 11:
```suggestion
```

The previous category `E-commerce` was fine. I don't think it fits very well in `Polls and events` since it's a sales/e-shop platform and `Polls and events` is for the organization/scheduling of events (see https://awesome-selfhosted.net/tags/polls-and-events.html)

nodiscc (changes requested):

nodiscc (approved):
Thanks

PR #651, opened 2024-04-05
Add platform to Erxes

ekcom:
Add Node.JS as a platform for erxes

### DIFF
--- a/software/erxes.yml
+++ b/software/erxes.yml
@@ -5,6 +5,7 @@ licenses:
   - GPL-3.0
 platforms:
   - Docker
+  - Nodejs
 tags:
   - Ticketing
 source_code_url: https://github.com/erxes/erxes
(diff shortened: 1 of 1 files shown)


nodiscc (approved):

software/erxes.yml
```diff
@@ -5,6 +5,7 @@ licenses:
   - GPL-3.0
```
nodiscc, line 5:
```suggestion
  - AGPL-3.0
  - ⊘ Proprietary
```

It seems the license has changed since this was first submitted:

> https://github.com/erxes/erxes/blob/master/LICENSE.md
> erxes is not permitted to be hosted as a SaaS version to compete with erxes Inc.

The project's description itself advertises it as "source available":

> Source available experience management infrastructure

You can replace `⊘ Proprietary` with `Commons-Clause` if you like, since this is very close to the Commons Clause (you cannot sell the software)

nodiscc (changes requested):

software/erxes.yml
ekcom: Commons clause seems reasonable to me

nodiscc (approved):
Thanks

PR #654, opened 2024-04-08
Update wikipedia link on URL shorteners

pmodin:
The heading was changed in revision on [11 September 2023](https://en.wikipedia.org/w/index.php?title=URL_shortening&oldid=1174944709) (without any update to its contents [since](https://en.wikipedia.org/w/index.php?diff=1214649442&oldid=1174944709&title=URL_shortening)). Additionally, update the text to reflect this change.


nodiscc (approved):
thank you

PR #662, opened 2024-04-14
Remove Pastr (unmaintained since Dec 2023, demo link down, no releases)

nodiscc:
This was added before the requirement of an initial release older than 4 months. Given that the project has only received 15 commits in all its history, and that the demo link is down (no such domain), let's just remove it.

- ref. #1


kokomo123 (approved):
LGTM, no issues

Ki-er (approved):
Considering it hasn't had anything since 4 months ago, may not get anything else added. All good here.

PR #712, opened 2024-05-20
Rename VanDAM to Manyfold, and update URLs

Floppy:
The project has changed name, and has a proper website, so I've updated the details :)


nodiscc (approved):
Thank you

PR #718, opened 2024-05-23
move Koillection to Inventory Management

nodiscc:


Rabenherz112:
I'm not sure if I would classify this as Inventory Management (though this is more of a gut feeling, and I don't have a strong opinion on the matter). It seems more designed for sharing collections rather than categorizing and tracking inventory.

nodiscc:
@Rabenherz112 you are right, even though it _can_ be used as an inventory management tool, it is more generic than that. Thanks for the review. Closing.

PR #740, opened 2024-06-06
doc: remove staticsitegenerators.net, update statigcgen.com link

nodiscc:
- https://github.com/bevry/staticsitegenerators-list is unmaintained since 7 months, last update checks no longer working
- no activity in https://github.com/bevry/staticsitegenerators-list/issues
- staticgen.com now redirects to jacmstack.org/generators

### DIFF
--- a/.github/ISSUE_TEMPLATE/addition.md
+++ b/.github/ISSUE_TEMPLATE/addition.md
@@ -49,7 +49,7 @@ related_software_url: "https://my.awesome.softwar.e/apps"
 To ensure your issue is dealt with swiftly, please check the following (check the boxes `[x]`):
 - [ ] Submit one item per pull issue. This eases reviewing and speeds up inclusion.
 - [ ] You have searched the repository for any relevant [issues](https://github.com/awesome-selfhosted/awesome-selfhosted-data/issues) or [PRs](https://github.com/awesome-selfhosted/awesome-selfhosted-data/pulls), including closed ones.
-- [ ] Any software you are adding is not already listed at any of [awesome-sysadmin](https://github.com/n1trux/awesome-sysadmin), [staticgen.com](https://www.staticgen.com/), [staticsitegenerators.net](https://staticsitegenerators.net/), [dbdb.io](https://dbdb.io/browse).
+- [ ] Any software you are adding is not already listed at any of [awesome-sysadmin](https://github.com/n1trux/awesome-sysadmin), [jamstack/static site generators](https://jamstack.org/generators/), [dbdb.io](https://dbdb.io/browse).
 - [ ] Any software project you are adding to the list is actively maintained.
 - [ ] Any software project you are adding was first released more than 4 months ago.
 - [ ] Any software project you are adding has working installation instructions.
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,7 +6,7 @@ Thanks for taking the time to suggest an addition to awesome-selfhosted!
 To ensure your Pull Request is dealt with swiftly, please check the following (check the boxes `[x]`):
 - [ ] Submit one item per pull request. This eases reviewing and speeds up inclusion.
 - [ ] You have searched the repository for any relevant [issues](https://github.com/awesome-selfhosted/awesome-selfhosted-data/issues) or [PRs](https://github.com/awesome-selfhosted/awesome-selfhosted-data/pulls), including closed ones.
-- [ ] Any software you are adding is not already listed at any of [awesome-sysadmin](https://github.com/n1trux/awesome-sysadmin), [staticgen.com](https://www.staticgen.com/), [staticsitegenerators.net](https://staticsitegenerators.net/), [dbdb.io](https://dbdb.io/browse).
+- [ ] Any software you are adding is not already listed at any of [awesome-sysadmin](https://github.com/n1trux/awesome-sysadmin), [jamstack/static site generators](https://jamstack.org/generators/), [dbdb.io](https://dbdb.io/browse).
 - [ ] The file you are adding is formatted as described in [addition.md](https://github.com/awesome-selfhosted/awesome-selfhosted-data/blob/master/.github/ISSUE_TEMPLATE/addition.md).
 - [ ] `Demo` links should only be used for interactive demos, i.e. not video demonstrations. 
 - [ ] Comments and unused optional fields have been removed.
--- a/tags/static-site-generators.yml
+++ b/tags/static-site-generators.yml
@@ -1,10 +1,8 @@
 name: Static Site Generators
 description: '[Static site generators](https://en.wikipedia.org/wiki/Web_template_system#Static_site_generators) generate full static HTML websites based on raw data, plain text files and a set of templates. '
 redirect:
-  - title: staticsitegenerators.net
-    url: https://staticsitegenerators.net
-  - title: staticgen.com
-    url: https://www.staticgen.com
+  - title: jamstack.org - Static Site Generators
+    url: https://jamstack.org/generators/
 related_tags:
   - Blogging Platforms
   - Photo and Video Galleries


kokomo123 (approved):
no problems with this, at least we have alternatives

Rabenherz112:
The new project you're linking to doesn't appear to be maintained anymore either. According to [their repository](https://github.com/jamstack/jamstack.org), the last commit was made 7 months ago. Additionally, their issue page is cluttered with spam (issues #948, #947, #929, #904, etc.).

Does it make sense to switch to a list that seems similarly unmaintained?

nodiscc:
Thank @Rabenherz112 for pointing out the lack of updates on https://github.com/jamstack/jamstack.org

So what would be the best course of action now?
1. leave one (or both) links untouched, knowing that the lists are unmaintained (but at least comprehensive)
2. Remove the delegation/redirect on the `Static Site Generators` tag, copy back to awesome-selfhosted those that are still relevant (even if we curate the list, there are thousands of them, this is no small task and more importantly will absolutely flood awesome-selfhosted with static site generators)
3. copy static site generators that are still relevant to `awesome-selfhosted-data`, set up our tooling to generate a separate list at https://github.com/awesome-selfhosted/awesome-static-site-generators + https://static-site-generators.awesome-selfhosted.net/
4. same as 3. but with a separate `awesome-static-site` generators
5. find a better maintained, external list of static site generators

I think 5. requires the less work (assuming such a list exists). 1. requires no work at all. 3. would be ideal as it would not overload the main list, and the work can be reused to better integrate https://github.com/awesome-foss/awesome-sysadmin with awesome-selfhosted (possibly generate the markdown list and a https://sysadmin.awesome-selfhosted.net, but it is still a decent amount of work... 4. would induce a lot of duplication in tooling/issue management, etc.

Rabenherz112:
I think we should first consider option 5; it seems to require minimal effort and could potentially yield good results.

As for the other options (2, 3, and 4), I believe you should decide on those since they will undoubtedly bring a lot of work, especially for you. From my perspective, option 1 is only a short-term solution because the list will become outdated and less helpful than just searching GitHub (listing generators that are either outdated or non-functional, etc..).

Also, I don't see how options 3 & 4 involve less work than option 2 other than not completely flooding the main readme/website.

Rabenherz112:
I looked into this issue last weekend (but completely forgot to actually write in this discussion :P).

Last week, the [bevry/staticsitegenerators-list](https://github.com/bevry/staticsitegenerators-list) was updated to include one new generator, though there were no other changes.

I also searched GitHub for alternative lists, which turned out to be more challenging than expected. Most "awesome-yourProgrammingLanguage" lists have a static generator section, but comprehensive lists that include all programming languages are rare. One of the few well-maintained and longer-standing ones is [myles/awesome-static-generators](https://github.com/myles/awesome-static-generators).

nodiscc:
> Last week, the [bevry/staticsitegenerators-list](https://github.com/bevry/staticsitegenerators-list) was updated to include one new generator

https://github.com/bevry/staticsitegenerators-list/commit/782848205a8069d65041e0071f9eeef16e062a8c but `aurora` cannot be found at https://staticsitegenerators.net/. Is their automatic generation broken? I think we should let them know (and link back here)

If there is a chance to get bevry/staticsitegenerators-list up and running again we should put the other solutions on hold

Rabenherz112:
Based on balupton's comment on the project's status in the mentioned issue, I think we can temporarily remove https://staticsitegenerators.net/. If the pipeline gets fixed and the project gains some ongoing maintenance, we can always re-add it later.

The Jamstack repo is active again, with recent commits starting from December 4, 2024, and has been relatively stable since. However, I'm unsure about the level of maintenance. There are still open spam/bot issues, along with quite a few PRs dating back to 2021.

@nodiscc, what are your thoughts?

nodiscc:
I think we should do the following:
- [ ] create empty `static-site-generators` and  `static-site-generators-html` repos under https://github.com/awesome-selfhosted/
- [ ] copy the CI tooling/pipeline from awesome-selfhosted-data to static-site-generators
- [ ] convert https://github.com/bevry/staticsitegenerators-list data to the YAML format used by our CI/tools

Once there we will have an equivalent of https://staticsitegenerators.net/, using the same tools/format as https://awesome-selfhosted.net/

- [ ] ask @balupton if he wants to deploy the resulting static website on his domain `staticsitegenerators.net`, and take ownership of the repo - basically `git fetch && git reset --hard master` on a cron job)
- [ ] or keep it under the awesome-selfhosted org and make him the primary maintainer, and establish the same community-maintained model as awesome-selfhosted

The only thing I won't personally do is curate it and review additions (too time consuming).

If he doesn't want to, we could host it under https://staticsitegenerators.awesome-selfhosted, but 1. I'm not too keen on hosting a website I don't maintain 2. we would lose already existing users and traction of https://staticsitegeneratos.net

Steps 1-3 should be easy enough

kokomo123:
As per comments above, I am dismissing this review for now 👍🏻

Rabenherz112:
Sounds like a solid plan. I’m happy to handle steps two and three if you create the repositories.

> and take ownership of the repo - basically git fetch && git reset --hard master on a cron job)

I might misunderstand something, but wouldn’t it make more sense to mirror what we do with the awesome-selfhosted repo? Converting the repo into a README (with balupton’s approval) to use its existing traffic? Or are you thinking about only hosting and maintaining the web version?

balupton:
I'm taken aback that at no point in this thread, with all the suggestions, plans, ideas, and efforts over a year - there was never the idea of contributing back, which would also solve the contention.

Why the desire to fork, why not support? The thousands of "top 10 static generators" articles all copying/forking each other in their own grifts, was why I created staticsitegenerators.net to consolidate their redundant efforts and deprecate their grifting competition. It's a pity such a desire to fork is happening again.

Re maintenance of the list, it is maintained by a group of people. There isn't changes that happen daily or weekly, but monthly. Regarding forking the list... why can't you just consume the list, why fork it?

Re maintenance of the website, getting it updated can be tracked and discussed here:
https://github.com/bevry/staticsitegenerators-website/issues/97 — Regarding forking and alternatives the website, that's fine, the project was designed where there would be a one-to-many relationship between the list and consumers of the list.

The ideal spirit of reciprocal open-source should remain, it is what created the value in these projects in the first place, it's the only spirit that can maintain a definitive list/effort, and PRs, maintenance, and assistance support for origins should be the first preference, to further unified impact and remove unnecessary redundancy.

It's great that this project is taking that spirit for self hosted data, which includes the subset segment of static site generators. The projects and efforts shouldn't need to compete with each other as far as I'm aware, and the self hosted superset of concerns could consume and support the subset of concerns that the static site generators listing/website projects are focused on.

nodiscc:
> there was never the idea of contributing back, which would also solve the contention

What I propose is essentially contributing, but completely replacing the build system/CI/data format with ours, would you agree with that?

The resulting website would look like https://awesome-selfhosted.net/ in the first time.

You would benefit from our existing tools that update projects metadata automatically, check for dead links and unmaintained projects, etc. It consists of a quite simple python program + a Makefile + Github actions.

The only remaining manual  maintenance task is reviewing PRs/additions, and pruning old projects when the CI emits warnings.

It will be faster for me to copy-paste out tooling in your repo, convert your data format tour YAML-based format - than to try to wrap my head around the existing system/CI and try to fix it - you said yourself in https://github.com/bevry/staticsitegenerators-list/issues/467 that it needed to be simplified/rewritten.

> remove unnecessary redundancy

That is exactly the point, since the beginning we consider https://staticsitegenerators.net/ a better resource for this specific category, and don't want to duplicate it, so we simply link to it (see https://awesome-selfhosted.net/tags/static-site-generators.html) - the only problem is the lack of updates caused by the broken tooling. The proposed solution would fix that.

We could work on replicating the existing "table" format in a second time, if you prefer that.

nodiscc:
Replaced by https://github.com/awesome-selfhosted/awesome-selfhosted-data/pull/1563

balupton:
staticsitegenerators.net / staticsitegenerators.bevry.me / https://github.com/bevry/staticsitegenerators now has an up to date listing

PR #742, opened 2024-06-08
Remove Kanbana (Unmaintained since June 2023)

Rabenherz112:
- ref: #1
- `ERROR:awesome_lint.py: Kanbana: last updated -367 days, 1:29:02.081263 ago, older than 365 days`

PR #745, opened 2024-06-09
remove kanbana (unmaintained since June 2023)

nodiscc:
- ref. #1
- `ERROR:awesome_lint.py: Kanbana: last updated -368 days, 1:29:26.072626 ago, older than 365 days`


Rabenherz112:
Duplicate to #742 ^^

PR #763, opened 2024-06-25
Delete software/partdb.yml duplicate with software/part-db.yml (#1)

nugrahayoga:
Delete the partdb.yml file because of the file generated file will be named as part-db.yml


nodiscc (approved):
Thanks

PR #765, opened 2024-07-04
Remove Seatsurfing demo

Rabenherz112:
- ref: #1
- `https://seatsurfing.app/get-started/ : HTTP 404`
- The Demo moved to `https://seatsurfing.app/getting-started`, however for the demo you now need to register a account (which also includes registering and confirming an email address).


nodiscc (approved):
Thanks

PR #766, opened 2024-07-04
fix hitobito website

Rabenherz112:
- ref: #1
- `https://hitobito.com/en : HTTP 404`
- Link to the domain without english suffix `https://hitobito.com`


nodiscc (approved):
Thanks

PR #767, opened 2024-07-04
Remove easyshortener

Rabenherz112:
- ref: #1
- `ERROR:awesome_lint.py: Easyshortener: last updated -366 days, 1:28:45.513551 ago, older than 365 days`


nodiscc (approved):
Thanks

PR #863, opened 2024-09-13
Delete kerberos.io NVR

konjas:
kerberos.io consists of 4 component. Here's a little summary of them:

- Agent
  - purpose: configures a single camera and saves recordings locally
  - price: free
  - sources: open
  - hosting: self-hosted
- Vault
  - purpose: centralized storage
  - price: free
  - sources: closed
  - hosting: self-hosted
- Factory
  - purpose: orchiestrates Agents on Kubernetes cluster
  - price: free
  - sources: closed
  - hosting: cloud
    - can be self-hosted if you pay at least 50 EUR/month for the Hub component
- Hub
  - purpose: manages the whole stack, including watching live streams and cameras PTZ control
  - price: 2-2000 EUR/month
  - sources: closed
  - hosting: cloud
    - can be self-hosted only if you pay 1000 EUR/month for the Hub component

While the Agent component is a standalone, free, open source, self-hosted piece of software, it's unfortunately limited to just setting up the camera and watching recordings.

Any other component seems to be optional, but without the Hub component, the user cannot watch the cameras streams live nor control them using PTZ. It leads to a conclusion that the Hub component, while theoretically optional, provides a core functionality, making it actually required in most use cases.

Since the Hub component can be self-hosted in only 1 out of 6 available pricing plans, which costs 1000 EUR/month, I would state that the whole kerberos.io is neither awesome nor self-hosted, thus should be removed from the list.


nodiscc (approved):
Thank you for taking the time to report/fix this.

Kerberos.io was added in https://github.com/awesome-selfhosted/awesome-selfhosted/pull/2444 and there was no thorough review of its licensing scheme/available self-hosting options at the time.

One could argue, that this would only mandate that we edit the entry to be about Kerberos.io Hub (since this is the central component) and mark it as proprietary, but given the very limited self-hosting options, and the existence of FOSS alternatives, let's just remove it for now. It can always be re-added/reviewed more carefully later.

kokomo123 (approved):
LGTM with the points above

PR #915, opened 2024-10-25
Remove viseron (duplicate entry)

AllieQpzm:
There's two entries for Viseron (`viseron.yaml` and `viseron.yml`), resulting it being dispayed twice under https://awesome-selfhosted.net/tags/video-surveillance.html

This PR removes `viseron.yaml`


nodiscc (approved):
Thank you

PR #916, opened 2024-10-26
remove LBRY (unmaintained since 2023-10-23)

nodiscc:
- ref #1
- ERROR:awesome_lint.py: LBRY: last updated -369 days, 1:26:07.877480 ago, older than 365 days


kokomo123 (approved):
No problems with this, LGTM.

PR #981, opened 2024-11-25
remove infinoted (unmaintained since Nov 2023)

Rabenherz112:
- ref: #1
- `ERROR:awesome_lint.py: Infinoted: last updated -366 days, 1:26:21.233157 ago, older than 365 days`


nodiscc (approved):
Thanks

PR #1007, opened 2024-12-13
remove osTicket (Unmaintained since Oct 2023)

Rabenherz112:
- ref: #1
- `ERROR:awesome_lint.py: osTicket: last updated -415 days, 1:22:42.573775 ago, older than 365 days`
- This is a well known project, but it seems like it is currently not maintained anymore. I found Issues asking about this as late as this September, but the response was that is still maintained (https://github.com/osTicket/osTicket/issues/6746). However looking at the roadmap linked to, which was last updated in May it doesn't seem like there is any activity. So not too sure about this one..


nodiscc (approved):
It is indeed strange, it used to be an active and widely used project. I will do some research, approving anyway, thanks.

Paolo590 (changes requested):
Aaaa

PR #1011, opened 2024-12-13
remove website of eda

Rabenherz112:
- ref: #1
- ` https://eda.jortilles.com/en/jortilles-english/ : HTTP 404`
- Website seems to have been removed, while the project is still active their is no link to the website

### DIFF
--- a/software/eda.yml
+++ b/software/eda.yml
@@ -1,5 +1,5 @@
 name: EDA
-website_url: https://eda.jortilles.com/en/jortilles-english/
+website_url: https://github.com/jortilles/EDA
 description: Web application for data analysis and visualization.
 licenses:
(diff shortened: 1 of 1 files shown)


nodiscc (approved):

Jnavar19:
Sir that not even me I just started this like 3 months ago yesterday I was on the site and I didn't know what I was doing and so I did ask for help and some men on there say collect the coins and then I collect the coins and then I was pressing buttons. I had no clue what I was doing, but like that name is not my name. Can you please tell me if we can have a talk because I don't want you to think I mean, I don't know what's going on. I don't know what I did. I was pressing button. That's why I asked. I need help cause I have no clue what I was doing.

PR #1031, opened 2024-12-21
remove duplicate bifroest

Rabenherz112:
```
Duplicate 'name' field detected: Engity's Bifröst
Files containing this name:
- D:\GIT\awesome-selfhosted-data\software\bifroest.yml (Created: 2024-12-16 22:53:06 +0000, Author: Gregor Noczinski)
- D:\GIT\awesome-selfhosted-data\software\engity's-bifröst.yml (Created: 2024-12-21 00:45:13 +0000, Author: awesome-selfhosted-bot)
```


Rabenherz112:
Also, why is there even an apostrophe in there, shouldn't this be removed?

nodiscc (changes requested):
Thank you (I really need to take some time to review/merge improvements to https://github.com/nodiscc/hecat/pulls by the way, I did not forget)

Both files have the same content but the file naming of `bifroest.yml` is more consistent than `engity's-bifröst.yml` (this one was imported automatically from the old markdown list), I'd rather delete the latter and keep `bifroest.yml`. Then its metadata will be updated automatically.

Rabenherz112:
> Both files have the same content but the file naming of bifroest.yml is more consistent than engity's-bifröst.yml (this one was imported automatically from the old markdown list), I'd rather delete the latter and keep bifroest.yml. Then its metadata will be updated automatically.

Are you sure this will work? I thought the processors/github_metadata had a bug with the `write_software_yaml` function. It takes the `name:` field for the file name and doesn't properly check if the current name of the YAML file is the same as the written file, which results in these duplicates in the first place.

So keeping `bifroest.yml` which has the wrong name: field, and deleting the auto-generated `engity's-bifröst.yml` wouldn't this just result in generating the `engity's-bifröst.yml` again next time the update metadata job runs?

nodiscc:
You are right! This is the cause of several duplicates (e.g. https://awesome-selfhosted.net/#scribble-rs)
I will update https://github.com/nodiscc/hecat to ensure that filenames are correctly formatted

nodiscc:
Fixed (alongside others + updated checks) in https://github.com/awesome-selfhosted/awesome-selfhosted-data/pull/1038

PR #1034, opened 2024-12-22
Add channeltube.yml

fubarpie:
Adding: ChannelTube - Download video or audio from YouTube channels on a schedule via yt-dlp.

To ensure your Pull Request is dealt with swiftly, please check the following (check the boxes `[x]`):
- [x] Submit one item per pull request. This eases reviewing and speeds up inclusion.
- [x] You have searched the repository for any relevant [issues](https://github.com/awesome-selfhosted/awesome-selfhosted-data/issues) or [PRs](https://github.com/awesome-selfhosted/awesome-selfhosted-data/pulls), including closed ones.
- [x] Any software you are adding is not already listed at any of [awesome-sysadmin](https://github.com/n1trux/awesome-sysadmin), [staticgen.com](https://www.staticgen.com/), [staticsitegenerators.net](https://staticsitegenerators.net/), [dbdb.io](https://dbdb.io/browse).
- [x] The file you are adding is formatted as described in [addition.md](https://github.com/awesome-selfhosted/awesome-selfhosted-data/blob/master/.github/ISSUE_TEMPLATE/addition.md).
- [x] `Demo` links should only be used for interactive demos, i.e. not video demonstrations. 
- [x] Comments and unused optional fields have been removed.
- [x] The file you are adding uses [kebab-case](https://en.wikipedia.org/wiki/Letter_case#Kebab_case) file naming, for example `my-awesome-software.yml`.
- [x] Values for `platform` are the main **server-side** requirements for the software. Don't include frameworks or specific dialects.
- [x] Any software project you are adding to the list is actively maintained.
- [x] Any software project you are adding was first released more than 4 months ago.
- [x] Any software project you are adding has working installation instructions.
- [x] You understand that your Pull Request will be merged ~1 week after approval, to allow for further comments if needed.

### DIFF
--- a/software/channeltube.yml
+++ b/software/channeltube.yml
@@ -1,0 +1,10 @@
+name: "ChannelTube"
+website_url: "https://github.com/TheWicklowWolf/ChannelTube"
+source_code_url: "https://github.com/TheWicklowWolf/ChannelTube"
+description: "Download video or audio from YouTube channels on a schedule via yt-dlp."
+licenses:
+  - AGPL-3.0
+platforms:
+  - Docker
+tags:
+  - Automation
--- a/software/channeltube.yml
+++ b/software/channeltube.yml
@@ -8,3 +8,4 @@ platforms:
   - Docker
 tags:
   - Automation
+depends_3rdparty: true


software/channeltube.yml
```diff
@@ -0,0 +1,10 @@
+name: "ChannelTube"
+website_url: "https://github.com/TheWicklowWolf/ChannelTube"
+source_code_url: "https://github.com/TheWicklowWolf/ChannelTube"
+description: "Download video or audio from YouTube channels on a schedule via yt-dlp."
+licenses:
+  - AGPL-3.0
+platforms:
+  - Docker
+tags:
+  - Automation
```
Ki-er, line 10: Depends on YT so need to add this.
```suggestion
  - Automation
depends_3rdparty: true
```

Ki-er (changes requested):

software/channeltube.yml
fubarpie: The template lists that field as optional.

TheGiraffe3: It's optional if false; if true the tag needs to be there.

fubarpie: That is not what the document says, and that is not what the word optional means. There are only two options, and if you have to include "true" if true, then it is not an optional field at all. The idea of a field being optional only if one of two options is selected is a paradox. 

In reality, the template simply declares this field as 'optional', and no further specification is made in the template or the contributing.md document. 

Furthermore, I would contest that the use of yt-dlp fulfills the definition of a 3rd party-dependency, since it can be used on thousands of video sites and other software that uses yt-dlp does not have this tag. The dev has shown in their blog how the software can be used to monitor any folder on your system for deletion, like a MeTube folder for example - making it useful regardless of the continued existence of Youtube. For this reason, I decided to leave this field blank as I don't believe it does depend on Youtube, but I did not want to declare it as false either just since so many users will likely want to use it primarily for Youtube channels.

nodiscc: I agree that proper use of the `depends_3rdparty` field is not very well explained.
If it's `false` it should be omitted entirely.
If the software depends on a 3rd-party, non-selfhostable network service (in this case YouTube), it must be set to `true`. 
yt-dlp itself is not the problem, does ChannelTube work with _any_ video hosting site supported by yt-dlp, or does it only work with YouTube channels?

fubarpie: It can be used to monitor any folder, so if you manually download files from any site using something like MeTube, it will monitor that folder and delete older videos. At this time the downloading of new videos is limited to YouTube. 

I'm not sure if that would make it considered to be dependent on YouTube - certainly the primary function that most users will want is to use it with YouTube, but I also personally use it to monitor several MeTube folders for auto-deletion which have videos from several platforms. So I know it can be used for more than just YouTube.

For me it was a bit of a gray area, and as the field is listed as optional, I choose to leave the field out. If the community feels that it should be included then I will also submit changes to pending pull request bookbounty.yml which relies on Library Genesis and for recently approved lidatube.yml, which uses yt-dlp and *does* only work with YouTube at the moment.

nodiscc: > feels that it should be included

I think it does, at least for now, and considering the project description and most of its documentation are geared toward usage with YouTube (things could be reconsidered if the documentation gave some example of usage with e.g. Peertube or another Free software video hosting tool)

> I will also submit changes to pending pull request bookbounty.yml which relies on Library Genesis

Yes, that would be correct

> lidatube.yml, which uses yt-dlp and does only work with YouTube

That would be great as well, thank you.

nodiscc: @fubarpie ping

nodiscc (approved):

PR #1044, opened 2024-12-29
Rename what-to-cook?.yml to what-to-cook.yml

Ki-er:
error: invalid path 'software/what-to-cook?.yml

when pulling from source, get this error, I think its from the '?'

For simplicity and to keep a standard name convention, I think we should remove it


kokomo123 (approved):
No issues here.

nodiscc:
This is caused by https://github.com/nodiscc/hecat/pull/143, the filename will be set to `what-to-cook?.yml` the next time the metadata update runs.
I will address this in a new pull request.

Ki-er:
Ah I didnt see the PR over at hecat. Ill close this for now then!

PR #1049, opened 2024-12-29
Remove Anuko (Unmaintained since December 2023)

Ki-er:
Remove Anuko 
`ERROR:awesome_lint.py: Anuko: last updated -367 days, 1:27:57.681123 ago, older than 365 days`
- Last commit since December 2023
- No releases


nodiscc (approved):
Thank you

PR #1054, opened 2024-12-29
Remove Github stats from Inboxen

Ki-er:
I think Inboxen used to be a github repo however the source code is now codeberg but it kept the GH stats. This broke the unmaintained projects action:

`ERROR:awesome_lint.py: Inboxen: last updated -379 days, 1:27:57.659409 ago, older than 365 days`


nodiscc (approved):
thanks

PR #1059, opened 2024-12-31
Add Elixir platform to Cannery app

shibaobun:
I forgot to add Elixir into the list of platforms for Cannery. While there aren't any host-specific installation instructions in the readme, the app should be straightforward to host like any Phoenix application.

Thank you!


Ki-er:
Only platforms with installation instructions should be listed.

nodiscc:
> Only platforms with installation instructions should be listed.

^ correct

PR #1083, opened 2025-01-04
Remove Kbin and associated link (Unmaintained since Dec 2023)

Ki-er:
No commits since 2023 
No releases at all.

Remove the kbin awesome selfhosted link as it seemed to be the same as rhe default kbin.social link which seems broken

other links such as https://kbin.pub/ do not work.  https://karab.in/ leads to nowhere.

This project overall seems inactive.


kokomo123 (approved):

nodiscc (approved):
thanks

PR #1089, opened 2025-01-08
Add Depends Flag to Youtube and Spotify Apps

Ki-er:
Added Depends flag to Youtube / Spotify apps.
- Lidatube: finding and fetching missing Lidarr albums via yt-dlp.
- Mopidy: plays music from local disk, Spotify, SoundCloud, Google Play Music, and more.
- Stretto:  video or song from youtube or soundcloud

Fixed description of Tube.yml


TheGiraffe3 (approved):

nodiscc:
Technically yt-dlp/youtube-dl can be used to download media from thousands of sites, some of which are actually Free Software network services (e.g Peertube instances). So I'm not sure we should apply the `depends_3rdparty` flag to youtbe-dl web interfaces and the like, **unless** they exclusively work on youtube or non-free services.

Rabenherz112:
I generally agree that it wouldn’t be fair to slap the `depends_3rdparty` flag on these wrappers just because they use youtube-dl, especially if went out of their way to also support free services (or just ensure that it will work with all supported sites).

> So I'm not sure we should apply the depends_3rdparty flag to youtbe-dl web interfaces and the like, unless they exclusively work on youtube or non-free services.

However, how do we check this though? If we go down this route, we’d have to review the code for every youtube-dl wrapper. For instance, PR #1034 uses youtube-dl, but if you look at the code (like at [this line](https://github.com/TheWicklowWolf/ChannelTube/blob/main/src/ChannelTube.py#L154 )), it’s really only set up for YouTube. Essentially, this means we’d need to comb through the source code of any submission that relies on youtube-dl. That would make our reviews suddenly a whole lot more complicated, time-consuming, and likely to introduce errors.

nodiscc:
> However, how do we check this

We only need to read the documentation/project description:
- https://github.com/alexta69/metube -> "Self-hosted YouTube downloader" -> `depends_3rdparty: true`
- https://github.com/nbr23/youtube-dl-server -> "Web / REST interface for downloading youtube videos" -> `depends_3rdparty: true`
- https://github.com/Totonyus/ydl_api_ng -> "Simple rest API for youtube-dl" -> `depends_3rdparty: false`

Maybe it's not accurate, but if the project description itself is not accurate, it's the only thing we can do.

software/ydl_api_ng.yml
```diff
@@ -8,6 +8,7 @@ platforms:
 tags:
   - Automation
 source_code_url: https://github.com/Totonyus/ydl_api_ng
+depends_3rdparty: true
```
nodiscc, line 11:
```suggestion
```

as per comment above

Ki-er:
Ill go through this PR and make some decisions on the above projects. Thanks!

software/youtubedl-material.yml
```diff
@@ -9,6 +9,7 @@ platforms:
 tags:
   - Automation
 source_code_url: https://github.com/Tzahi12345/YoutubeDL-Material
+depends_3rdparty: true
```
nodiscc, line 12:
```suggestion
```

"a Material Design frontend for [youtube-dl](https://rg3.github.io/youtube-dl/)." The README also mentions downloading from other supported sites

nodiscc (changes requested):

Ki-er:
Updated the PR and left project quotes in the description.

nodiscc (approved):

PR #1107, opened 2025-01-15
Fix Papermerge (New GH Link)

Ki-er:
No commits since Jan 2024
No releases since April 2024

Placing this on hold until next week as communication within https://github.com/ciur/papermerge/issues/643 and https://github.com/ciur/papermerge/issues/641 suggest this may be active, however if no commits or releases next week, then we can merge this.


ciur:
Hi,

papermerge author here.
Papermerge project is under very active development. 
The misleading point here is the repository. Repository where most of the development happens is:

[https://github.com/papermerge/papermerge-core](https://github.com/papermerge/papermerge-core)

There are many more, all under https://github.com/papermerge/ github org.

The misleading part here is that, indeed, I've started project in https://github.com/ciur/papermerge, but later, I've split it into multiple repos, under papermerge github org. The repository which you mentioned is there because of:
- "historical reasons"
- also I use it for tickets

Anyway, just wanted to clear misunderstanding that papermerge is not being maintenaned.

Screenshot from 2025-01-15 20-21-12

Ki-er:
Thanks @ciur 

I have made the following changes:
- Changed GH Link to maintained version
- Cleaned up description
- Removed python platform as there are no installation instructions that I could find.

nodiscc (approved):

PR #1109, opened 2025-01-15
Add more URLs to link checker / Fix URLs

Ki-er:
Added more url links to the url check file

I also fixed a couple redirect links that were better to fix in the yml than add them to the link checker


nodiscc (approved):
Thank you
Out of curiosity I checked these URLs with `curl -L --head $URL` and they all return the `server: cloudflare` header

PR #1113, opened 2025-01-21
fix: correct license and platforms for Manticore Search

sanikolaev:
Manticore is now GPL-v3 licensed (switched from GPL-v2 a while ago). There's also a Helm Chart to run it in K8S - https://github.com/manticoresoftware/manticoresearch-helm


Rabenherz112 (commented):
The License change seems fine, but I couldn't find a way to install Manticore via K8S in the [install docs.](https://manual.manticoresearch.com/Installation/Installation) Where is this documented?

nodiscc (approved):

PR #1119, opened 2025-01-28
header: clearly mark the markdown version as legacy/deprecated

nodiscc:
I saw on several occasions people still linking to the old, single flat markdown file version of the list. This does not appear to be out of preference for the old format, but rather not knowing about the "new" (1.5 year old already) website.

This PR adds a rather obvious banner at the top of https://github.com/awesome-selfhosted/awesome-selfhosted (not on the website)

image

### DIFF
--- a/.hecat/export-markdown.yml
+++ b/.hecat/export-markdown.yml
@@ -5,7 +5,7 @@ steps:
       source_directory: ./
       output_directory: awesome-selfhosted
       output_file: README.md
-      markdown_header: markdown/header.md
+      markdown_header: markdown/header-singlepage.md
       markdown_footer: markdown/footer.md
       back_to_top_url: '#awesome-selfhosted'
       exclude_licenses:
--- a/markdown/header-singlepage.md
+++ b/markdown/header-singlepage.md
@@ -6,4 +6,7 @@ Self-hosting is the practice of hosting and managing applications on your own se
 
 This is a list of [Free](https://en.wikipedia.org/wiki/Free_software) Software [network services](https://en.wikipedia.org/wiki/Network_service) and [web applications](https://en.wikipedia.org/wiki/Web_application) which can be hosted on your own server(s). Non-Free software is listed on the [Non-Free](https://github.com/awesome-selfhosted/awesome-selfhosted/blob/master/non-free.md) page.
 
+> [!IMPORTANT]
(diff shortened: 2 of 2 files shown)


Rabenherz112 (approved):
I think the main reason people still link to the "old" version of this list is simply that they prefer it over the web version (myself included).

That said, I don't see any reason why we shouldn't make it clearer that this version is deprecated.

nodiscc:
> they prefer it over the web version (myself included).

That is interesting feedback, I don't think we ever discussed this before. What are the reasons?

Rabenherz112:
Hmm, this is much more difficult to put into words than I thought... I think I can break this down into three main reasons why I still prefer the Markdown version over the website

**Compact Information**
I’m a big fan of compact information; the more I can see at once, the better. On GitHub, I can usually see an entire category on my screen, while on the website, I’m stuck with +/- 5 software entries. That wouldn't be a huge deal if the site gave me more useful details per entry, but right now, it doesn’t (see below).

**The UI Feels Cluttered**
Scrolling through the website feels a bit overwhelming. There are a lot of colors, tags, buttons, and labels packed closely together, all the same size and right next to each other. It’s not always clear what’s a button, which elements have tooltips, and which don’t. In contrast, the Markdown version is clean and straightforward—just a simple list that’s easy to navigate.

Also, I’m not really a fan of the overall design (not saying I could do it better, so no offence meant). It just feels a bit outdated and not as smooth as it could be.

**No Real Advantage Over GitHub**
Since the Markdown list came first, the website should make things easier or better; but it doesn’t for me.
- The built-in search is pretty bad (no hate, I wouldn’t be able to do it better either). If I want to find all Docker apps and search exactly for "docker", the search will not display a single software that works with Docker _(seriously, try it)_. Even if I know the exact software name (why would I even use the search then) and the software name is unique, I still get at least 4 results. Sure, I can use CTRL+F on the website (which honestly works better than the built-in search), but I can do that in GitHub too; so there’s no real advantage here. And something like “show me all Pastebins that can be deployed with Docker” isn’t possible yet.
- The website does show star counts and last update dates, which I can’t see in the Markdown version (an advantage, nice). But since I can’t sort by them or do anything useful with that info, it doesn’t really add much value for me.

When I want to use a web version (which happens rarely) I mostly use third-party tools such as [awwesome](https://awweso.me/) or the new [ASH UI](https://awesomeselfhosted.netlify.app/catalog?search=Docker&category=Bookmarks+and+Link+Sharing).

PR #1131, opened 2025-02-02
remove inlets (unmaintained since jan 2024)

Rabenherz112:
- ref: #1
- `ERROR:awesome_lint.py: inlets: last updated -370 days, 1:27:54.250331 ago, older than 365 days`


nodiscc (approved):
thanks

PR #1195, opened 2025-03-11
Remove Fileshelter (Unmaintaied since March 2024)

Ki-er:
As part of #1
`ERROR:awesome_lint.py: FileShelter: last updated -372 days, 1:25:42.878969 ago, older than 365 days`


nodiscc (changes requested):
@Ki-er you didn't actually delete the file, just emptied it

Ki-er:
Should be fixed!

nodiscc (approved):
Thanks!

PR #1203, opened 2025-03-15
Remove Apache HTML Server (Unmaintained Since 2017, Dead Source Code)

Ki-er:
As part of #1
`viewvc 2017-01-28 17:49  2.0K Web view of SVN repositories`

`https://svn.apache.org/viewvc/httpd/httpd/branches/2.4.x/ : HTTPSConnectionPool(host='svn.apache.org', port=443): Read timed out. (read timeout=10)`

Forbidden resource

PR #1204, opened 2025-03-15
Fix source code url for apache http server

kokomo123:
This is for simply adjusting apache's HTTP server source code to the mirror referenced in #1203 thus avoiding having to outright delete the server entirely from the list. I'm open to feedback from collaborators if this isn't the clear cut solution

### DIFF
--- a/software/apache-http-server.yml
+++ b/software/apache-http-server.yml
@@ -1,6 +1,6 @@
 name: "Apache HTTP Server"
 website_url: "https://httpd.apache.org/"
-source_code_url: "https://github.com/apache/httpd"
+source_code_url: "https://svn.apache.org/repos/asf/httpd/httpd/trunk/"
 description: "Secure, efficient and extensible server that provides HTTP services in sync with the current HTTP standards."
 licenses:
   - Apache-2.0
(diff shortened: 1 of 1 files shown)


software/apache-http-server.yml
```diff
@@ -1,6 +1,6 @@
 name: "Apache HTTP Server"
 website_url: "https://httpd.apache.org/"
-source_code_url: "https://svn.apache.org/viewvc/httpd/httpd/branches/2.4.x/"
+source_code_url: "https://github.com/apache/httpd"
```
nodiscc, line 3: The correct URL is https://svn.apache.org/repos/asf/httpd/httpd/trunk/ (https://httpd.apache.org/dev/devnotes.html)

The github project is just a mirror

nodiscc (changes requested):

software/apache-http-server.yml
kokomo123: I see i see then 👍🏻 I just wanted to future proof it a little in case the url breaks again (which hopefully does not in this case)

Rabenherz112 (approved):

nodiscc (approved):

PR #1209, opened 2025-03-18
Remove Hackershare (Unmaintained since March 2024)

Ki-er:
As part of #1
`ERROR:awesome_lint.py: Hackershare: last updated -366 days, 1:24:31.245725 ago, older than 365 days`

I cant seem to load the site either and CI failing


kokomo123 (approved):
no problems with this

PR #1237, opened 2025-03-26
Add Feature Toggle Tag

Ki-er:
In an effort to clean up the misc section, I found two feature toggle options that could be moved to their own category.  I added Posthog too.

I found featbit, see other PR #1238 , to add to this category


kokomo123 (approved):
Think this would be nice to have, no problem having another tag here. LGTM, pending other thoughts from collaborators

Rabenherz112 (commented):
Great idea! Though isn't "Feature Flag" the more mainstream term instead of "Feature Toggle"?

Ki-er:
I had originally named it Feature Flag but Feature Toggle is the wiki definition so stuck with that.

nodiscc (approved):
Excellent idea

PR #1298, opened 2025-04-13
fix description of Bubo Reader

Rabenherz112:
Fixed current build error


software/bubo-reader.yml
```diff
@@ -1,6 +1,6 @@
 name: Bubo Reader
 website_url: https://github.com/georgemandis/bubo-rss
-description: `"Irrationally minimal" RSS feed reader.`
+description: Irrationally minimal" RSS feed reader.
```
Ki-er, line 3:
```suggestion
description: Irrationally minimal RSS feed reader.
```

Ki-er (changes requested):
I tried to keep the "" but must have broke it!

PR #1317, opened 2025-04-19
Remove Dat Project (Unmaintained organisation)

Ki-er:
As part of #1

So this is a long winded one. The organisation linked in the file says `[ DEPRECATED - More info on active projects and modules at https://dat-ecosystem.org/ ]`.

So you head to https://dat-ecosystem.org/ and find various projects on GH with a new [`Dat Ecosystem`](https://github.com/dat-ecosystem) however alot of these projects are 10 months old at a mininum [1](https://github.com/dat-ecosystem/dat-garden-rake) [2](https://github.com/playproject-io/datashell) or just dont exist - see `Ahau` on the site. Alot of other GH orgs are also linked [3](https://github.com/cabal-club/cable) [4](https://github.com/AraBlocks/ara-file-manager) with just as dead projects.

Overall I think that this should be removed and the individual projects should be added but to be honest, they all seemed unmaintained from the few ones I investigated.


kokomo123 (approved):
LGTM, no issues here. I agree with your sentiments.

nodiscc (approved):
Thank you
Agree with your conclusions

PR #1318, opened 2025-04-19
General clean up of projects beginning with D, E and F

Ki-er:
Part of the series of updating projects.

I combined D through to F since D and E didnt have too many changes.

Updated descriptions
Other changes include:
`Dify.ai` - Remove empty space
`Discount Bandit` - Demo leads to 404
`ERPNext` - Updated website link
`Freescout` - Update main link, Added demo
`FreshRSS` - Related software is archived

PR #1322, opened 2025-04-21
Remove EasyRSS as related software

JackStuart:

[EasyRSS](https://github.com/Alkarex/EasyRSS) has been archived as off 28/11/24


Ki-er:
This will be removed in #1318

PR #1353, opened 2025-05-07
Remove Readymedia (Unmaintained since May 2023)

Ki-er:
As part of #1
No commits since 2023
Cant find a replacement repo on GitHub


Ki-er:
There is a lot of open bugs found here: https://sourceforge.net/p/minidlna/bugs/

 Including a potential security vulnerability (?) https://sourceforge.net/p/minidlna/bugs/364/

 and a issue with the web interface that seems to be a couple year old https://sourceforge.net/p/minidlna/bugs/346/ 

I haven't used this software, it was the lack of commits that drew me to the removal but after doing some digging, it definitely seems unmaintained based on the number of open ended issues. Unfortunately since its not on GitHub, I couldn't get `trivy` to run to check for vulnerabilities. 

If others warrant this to stay, we can close this PR. Since its not on GitHub, we wont need to add it `last_updated_skip`.

nodiscc:
The supposed security vulnerability is a non-issue https://security-tracker.debian.org/tracker/CVE-2024-51442.
All bugs in the bug tracker seem rather minor.
However there are similar, probably equivalent solutions (Rygel) already in the list, so removing it Readymedia/MiniDLNA is probably fine, pending further reviews.

Rabenherz112 (approved):
I support removing it. The last update was two years ago, and with all the unresolved minor issues, it can be considered abandonware. Since we have other actively maintained projects serving the same use case, there’s no reason to keep it on the list.

PR #1363, opened 2025-05-10
Create Health and Fitness Tag

Ki-er:
Create a health and fitness category. 

Im not a huge fan of the description so happy to workshop some ideas but with the addition of alot of these projects recetly, thought this would be a good addition. 

Geneaology seemed like a good related tag to have as in my mind geneology is related to health but happy to discuss.

This will also update #1351 if this is merged.


kokomo123 (approved):
It makes sense for a new category for sure as we have enough software to categorize it like this as health related apps should belong in this instead of misc. LGTM, pending further thoughts from collaborators.

Rabenherz112 (approved):
Since you didn't like the current description that much, here an alternative option (not really sure, how much better it is, maybe use it as an inspiration):
```
[Electronic health records](https://en.wikipedia.org/wiki/Electronic_health_record), [Medical](https://en.wikipedia.org/wiki/Medical_software) and [Fitness trackers](https://en.wikipedia.org/wiki/Fitness_tracker) applications covering medical data management, personal wellness, and activity monitoring.
```

Also somehow in my head Genealogy and Health / Fitness don't fit together. Though if the others are fine with it, I also have no issues there.

Otherwise going ahead and approving it.

tags/health-and-fitness.yml
```diff
@@ -0,0 +1,4 @@
+name: Health and Fitness
+description: '[Medical](https://en.wikipedia.org/wiki/Medical_software), [Health](https://en.wikipedia.org/wiki/Health_information_technology) and [Fitness](https://en.wikipedia.org/wiki/Fitness_tracker) software.'
+related_tags:
+ - Genealogy
```
nodiscc, line 4:
```suggestion
```

I also don't find this relation relevant. Other than that, good to merge, thanks

nodiscc (changes requested):

nodiscc (approved):

Ki-er:
I remembered that discussion but couldnt find it again for reference. Thanks!

PR #1379, opened 2025-05-17
update inventaire source code URL

nodiscc:
- https://github.com/inventaire/inventaire/commit/e12f2de6bd85853a7759cf576d8633994a4de05e


twtulio (changes requested):
@

PR #1402, opened 2025-05-26
Remove Revolt (Invalid for awesome-selfhosted)

Ki-er:
As part of #1

> What does not qualify
> Software that depends on a specific cloud provider

Software instructions focus heavily on Hostinger. There is some commands at the bottom but without any other context outside of Hostinger,  especially in configuring the custom domain, people may get confused.


Rabenherz112 (commented):
> Software instructions focus heavily on Hostinger

While it is true, that the opening sections read more like an advertisement for Hostinger’s VPS solution. If you scroll down to the part beginning `After installation, SSH into the machine:`, it simply covers the program/VPS setup, which isn’t inherently problematic and does not depend on a specific cloud provider, but just requires a linux vm.

Although the domain configuration steps use Hostinger's DNS interface, that could easily be swapped out for any generic provider.

> There is some commands at the bottom but without any other context

As for the "Advanced Deployment" section, it simply walks you through the requisite Git and Docker commands needed to pull and run the application. Nothing here is specific to Hostinger, and the commands themselves are self-explanatory. The only potential stumbling block is in the additional Notes section, as they assume you already have Revolt up and running, which could be clearer, but otherwise there's no real issue.

---

I was able to spin up a Docker deployment on my own server in a few minutes, so there’s no dependency on any particular cloud provider.

chrisbeardy:
The first sentence on the readme deployment section states "The instructions going forward will use Hostinger as an example hosting platform, but you should be able to adapt these to other platforms as necessary"

nodiscc:
Thanks everyone for investigating this.
Based on these comments, I don't think removing it is warranted.

PR #1404, opened 2025-05-26
Delete software/rustdesk.yml

jameskimmel:
Unfortunately no longer really "selfhosted". 

Sure you can spin up a docker, but there is no way to configure the client to actually make use of your selfhosted server, unless you pay 20$ a month to get the client generator.


Ki-er:
Do the other platforms allow for editing of the config. If so it may be better to remove Docker but still leave the project on the list.

nodiscc:
> manually copy pasting it into the client

Does this work? Is this documented? Is it actually blocking/crippling the software?

> get banned for asking about selfhosting in the subreddit /r/rustdesk

Any evidence of this?

jameskimmel:
> Does this work? Is this documented? Is it actually blocking/crippling the software

It is crippling the software. Imagine you do remote support for a customer.
You would have to guide through that customer to the right settings and then have to enter the server TLD plus a public key.

> Any evidence of this?
https://old.reddit.com/r/rustdesk/comments/1jgexwu/and_now/
https://old.reddit.com/r/rustdesk/comments/1kvprqe/client_deployment_for_free_version/
https://old.reddit.com/r/selfhosted/comments/1kvqil8/selfhost_rustdesk/
https://old.reddit.com/r/rustdesk/comments/1kslbjo/do_i_need_to_buy_basic/

kokomo123 (approved):
Approving as it's been a while, with my comments above.

jameskimmel:
https://github.com/awesome-selfhosted/awesome-selfhosted-data/pull/1455

PR #1421, opened 2025-06-03
added archiving tag to paperless

chrisbeardy:
paperless itself says its an archiving software on their repo about "A community-supported supercharged version of paperless: scan, index and archive all your physical documents"

### DIFF
--- a/software/paperless-ngx.yml
+++ b/software/paperless-ngx.yml
@@ -8,6 +8,7 @@ platforms:
   - Docker
 tags:
   - Document Management
+  - Archiving and Digital Preservation (DP)
 source_code_url: https://github.com/paperless-ngx/paperless-ngx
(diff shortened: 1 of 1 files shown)


nodiscc:
I don't think it really fits the tags description https://github.com/awesome-selfhosted/awesome-selfhosted-data/blob/master/tags/archiving-and-digital-preservation-dp.yml

Software listed in https://awesome-selfhosted.net/tags/archiving-and-digital-preservation-dp.html is quite different.

chrisbeardy:
The tag is mostly filled with web preservation tools however there are physical media tools too such as Collective Access - Providence. One of the use cases of paperless as described by themselves is to archive your physical documents which is referred to in the archiving Wikipedia link.

PR #1450, opened 2025-06-18
Remove Speed Test by OpenSpeedTest™ (Unmaintained since June 2024)

Ki-er:
Part of #1
`ERROR:awesome_lint.py: Speed Test by OpenSpeedTest™: last updated -379 days, 1:21:15.959426 ago, older than 365 days`


kokomo123:
I agree with raben here, I added the discussion label so we could talk about it for now as i'm open to others thoughts 👍🏻

PR #1465, opened 2025-06-30
Exclude OpenSpeedTest from last updated check

Rabenherz112:
As mentioned in #1450, this RP adds OpenSpeedTest™ to `last_updated_skip` instead of removing it.


Ki-er:
I think we need to add this to https://github.com/awesome-selfhosted/awesome-selfhosted-data/blob/master/.hecat/awesome-lint-strict.yml aswell.

Rabenherz112:
You are absolutely correct, thanks.
(ignore the check error, that comes from #1466)

nodiscc (approved):
thanks

PR #1476, opened 2025-07-05
Remove Dnote (Unmaintained since July 2024)

Ki-er:
As part of #1 
Removes Dnote


Rabenherz112:
> without any external dependencies 

https://github.com/dnote/dnote/blob/master/go.mod
The project actually pulls in quite a few external modules.
```bash
trivy repo https://github.com/dnote/dnote --scanners vuln
# go.mod (gomod)
# Total: 6 (UNKNOWN: 0, LOW: 0, MEDIUM: 3, HIGH: 2, CRITICAL: 1)
```
But none of them appear to be exploitable or relevant in this context.

nodiscc:
I don't think we should exclude it from the checks.
If it has external dependencies they should be kept somewhat up-to-date (or at least not be affected by a CRITICAL vulnerability, even if it's not exploitable).
Also the list already contains a few similar alternatives. I think we should remove it.

kokomo123 (approved):
LGTM, no problems here.

nodiscc (approved):

chinofor2-glitch (approved):

PR #1477, opened 2025-07-06
Update website_url in django-crm.yml

DjangoCRM:
Add the new Django CRM website
https://github.com/DjangoCRM/django-crm


Rabenherz112 (approved):
Thanks

PR #1480, opened 2025-07-07
Add Wallabag to 'Bookmarks' category

PassionateBytes:
Wallabag belongs not just to the `Archiving & DP` category, but should appear under `Bookmarks and Link Sharing` as well, as that is a very common use case for the application.

### DIFF
--- a/software/wallabag.yml
+++ b/software/wallabag.yml
@@ -7,6 +7,7 @@ platforms:
   - PHP
 tags:
   - Archiving and Digital Preservation (DP)
+  - Bookmarks and Link Sharing
 source_code_url: https://github.com/wallabag/wallabag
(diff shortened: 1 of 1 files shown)


nodiscc:
I don't see any mention of sharing in the official documentation. As far as I know Wallabag is a purely personal, private read-it-later application.

PassionateBytes:
Fair point, but doesn't that fall under the Bookmark category? Having it under archiving just doesn't entirely make sense to me... 🤔

Rabenherz112:
I believe the current category is appropriate. By design, it’s a “read later” application that saves articles to your local storage, making it a perfect fit for the Archiving and Digital Preservation category.

PR #1495, opened 2025-07-18
Updated chhoto-url.yml

SinTan1729:
Added links to demo and related software.

### DIFF
--- a/software/chhoto-url.yml
+++ b/software/chhoto-url.yml
@@ -12,3 +12,5 @@ source_code_url: https://github.com/SinTan1729/chhoto-url
 stargazers_count: 313
 updated_at: '2025-07-17'
 archived: false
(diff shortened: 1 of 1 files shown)


software/chhoto-url.yml
```diff
@@ -12,3 +12,5 @@ source_code_url: https://github.com/SinTan1729/chhoto-url
 stargazers_count: 313
 updated_at: '2025-07-17'
 archived: false
+demo_url: "https://chhoto-url-demo.sayantansantra.com"
```
Rabenherz112: Demo requires Password which is only provided here
```suggestion
demo_url: "https://github.com/SinTan1729/chhoto-url?tab=readme-ov-file#demo"
```

SinTan1729: Thanks. Fixed it.

Rabenherz112 (approved):

PR #1507, opened 2025-07-26
Remove Maybe from the list.

kokomo123:
It seems Maybe has been officially discontinued and will no longer be maintained, and have a up to date version of the software. the company has pivoted per https://github.com/maybe-finance/maybe/releases/tag/v0.6.0. There is a hopeful successor named https://github.com/we-promise/sure respectively, but we need to give it some time to mature.


nodiscc (approved):
Thanks

PR #1557, opened 2025-07-30
Remove homebox (unmaintained since Jul 2024)

nodiscc:
- `ERROR:awesome_lint.py: homebox: last updated -379 days, 1:18:25.309412 ago, older than 365 days`


Rabenherz112 (approved):
We already have https://github.com/sysadminsmedia/homebox (the successor of this)

PR #1575, opened 2025-08-10
fix alternative to description

Rabenherz112:
This PR fixes some software description, which use the "alternative to" wording, but not using the intended format


kokomo123 (approved):
thanks for the fix!

PR #1587, opened 2025-08-17
fix screenconnect url

Rabenherz112:
- ref: #1
- `https://www.connectwise.com/platform/unified-management/control : HTTP 404`
- The website for ScreenConnect has been removed, Kokomo found that the product now has an extra domain


kokomo123 (approved):
LGTM 👍🏻 thanks for the fix!

PR #1605, opened 2025-08-26
Update tiny-tiny-rss

xuars:
Updated demo link and source code link (changed to gitlab).

### DIFF
--- a/software/tiny-tiny-rss.yml
+++ b/software/tiny-tiny-rss.yml
@@ -8,5 +8,5 @@ platforms:
   - PHP
 tags:
   - Feed Readers
-source_code_url: https://git.tt-rss.org/fox/tt-rss
-demo_url: https://srv.tt-rss.org/tt-rss/
+source_code_url: https://gitlab.tt-rss.org/tt-rss/tt-rss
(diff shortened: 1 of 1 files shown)


software/tiny-tiny-rss.yml
```diff
@@ -8,5 +8,5 @@ platforms:
   - PHP
 tags:
   - Feed Readers
-source_code_url: https://git.tt-rss.org/fox/tt-rss
-demo_url: https://srv.tt-rss.org/tt-rss/
+source_code_url: https://gitlab.tt-rss.org/tt-rss/tt-rss
```
nodiscc: I don't see any mention of this being the canonical source code URL. https://git.tt-rss.org/fox/tt-rss.git/log/ is still there and does not mention it. As far as I know it is used to allow a more "modern" contribution workflow/forks, but the "official" source code URL is still https://git.tt-rss.org/fox/tt-rss

Any more information?

xuars: > Any more information?

No, I thought it would be better to use the  gitlab one, but it doesn't really matter.

nodiscc (approved):
thanks

Issue #1608, opened 2025-08-28
Use Github Releases to publish merged json via automated github action based on yq


jo-chemla:
Following discussion in thread https://github.com/awesome-selfhosted/awesome-selfhosted-data/issues/1479#issuecomment-3228570951 regarding awesome-selfhosted alternative frontends and their data-source, it might be useful to use github releases as a way to provide users with a dump of the data on every (or tag selection) commit/pr (or on a tag selection like every month etc). 

Automating the processing of individual yaml files into a single json (+ single yaml ?) on every awesome-selfhosted release, via the [yq github-action](https://github.com/marketplace/actions/yq-portable-yaml-processor) could be the way to go.

Example using yq, see the [merging-all-files doc](https://mikefarah.gitbook.io/yq/operators/multiply-merge#merging-all-files)
```bash

# works in linux and github workflow
yq -o=json -I 0 ea '[.] as $item ireduce ([]; . *+ $item )' *.yml

# works in powershell
yq -o=json -I 0 ea '[.] as $item ireduce ([]; . *+ $item )' (Get-ChildItem -Name) > softwares.json

```

Combining this with the [softprops/action-gh-release](https://github.com/softprops/action-gh-release) action, it would automate the yq processing and be part of the release.

jo-chemla:
Just made a PR here that do look like a great starting point! https://github.com/awesome-selfhosted/awesome-selfhosted-data/pull/1609

The whole workflow runs in 20s, results in the following kind of release [0.1.4](https://github.com/jo-chemla/awesome-selfhosted-data/releases/tag/0.1.4) with resulting [software.json](https://github.com/jo-chemla/awesome-selfhosted-data/releases/download/0.1.4/software.json). 

Some notes mentioned in PR: 
- file-size: 492KB without indentation (one-line json), 670KB with 2-spaces indentation and simplifies reading a lot so used this.
- JSON object root is `{software: [yaml_list_transpiled_to_json}` to make it a valid json - could not be an array at root level.

nodiscc:
See comments in https://github.com/awesome-selfhosted/awesome-selfhosted-data/pull/1609

jo-chemla:
Thanks for the comments. For anyone stumbling on this thread instead of the linked PR, I'm duplicating some comments I replied to here https://github.com/awesome-selfhosted/awesome-selfhosted-data/pull/1609#issuecomment-3325410085

[...]

PR #1595, opened 2025-08-20
Add alternative frontends to footer.md

jo-chemla:
Resolves https://github.com/awesome-selfhosted/awesome-selfhosted-data/issues/1479

Thanks for taking the time to suggest an addition to awesome-selfhosted! 

To ensure your Pull Request is dealt with swiftly, please check the following (check the boxes `[x]`):
- [x] Submit one item per pull request. This eases reviewing and speeds up inclusion.
- [x] You have searched the repository for any relevant [issues](https://github.com/awesome-selfhosted/awesome-selfhosted-data/issues) or [PRs](https://github.com/awesome-selfhosted/awesome-selfhosted-data/pulls), including closed ones.
- [ ] Any software you are adding is not already listed at any of [awesome-sysadmin](https://github.com/awesome-foss/awesome-sysadmin), [staticgen.com](https://www.staticgen.com/), [staticsitegenerators.bevry.me](https://staticsitegenerators.bevry.me/), [dbdb.io](https://dbdb.io/browse).
- [ ] The file you are adding is formatted as described in [addition.md](https://github.com/awesome-selfhosted/awesome-selfhosted-data/blob/master/.github/ISSUE_TEMPLATE/addition.md).
- [ ] `Demo` links should only be used for interactive demos, i.e. not video demonstrations. If login credentials are required to access the demo, please link to the credentials directly.
- [ ] Comments and unused optional fields have been removed.
- [ ] The file you are adding uses [kebab-case](https://en.wikipedia.org/wiki/Letter_case#Kebab_case) file naming, for example `my-awesome-software.yml`.
- [ ] Values for `platform` are the main **server-side** requirements for the software. Don't include frameworks or specific dialects.
- [ ] Any software project you are adding to the list is actively maintained.
- [ ] Any software project you are adding was first released more than 4 months ago.
- [ ] Any software project you are adding has working installation instructions.
- [x] You understand that your Pull Request will be merged ~1 week after approval, to allow for further comments if needed.

PR #1609, opened 2025-08-28
Publish merged json via automated github action based on yq

jo-chemla:
Following https://github.com/awesome-selfhosted/awesome-selfhosted-data/issues/1608 here is a github action that relies on yq to push a json build on release. 

Some notes
 - Workflow Runs in less than 10s
 - file-size: 492KB without indentation (one-line json), 670KB with 2-spaces indentation and simplifies reading a lot so used this. 
 - JSON object root is `{software: [yaml_list_transpiled_to_json}` to make it a valid json - could not be an array at root level.
 - Example release here https://github.com/jo-chemla/awesome-selfhosted-data/releases/tag/0.1.4 and [software.json](https://github.com/jo-chemla/awesome-selfhosted-data/releases/download/0.1.4/software.json) file

### DIFF
--- a/.github/workflows/release-awesome-builds.yml
+++ b/.github/workflows/release-awesome-builds.yml
@@ -1,4 +1,4 @@
-name: Release awesome builds
+name: Release JSON for awesome-data builds
 
 on: 
   push:
@@ -10,12 +10,11 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v5
-      - name: Get an entry with a variable that might contain dots or spaces
-        id: get_username
+      - name: Merge software/*.yml into software.json via yq
         uses: mikefarah/yq@master
         with:
           cmd: yq ea '[.] as $item ireduce ([]; . *+ $item )' software/*.yml | yq -o=json -I 0 '{\"softwares\":.}' > software.json
-      - name: Test
+      - name: Raw dump software.json
         run: cat software.json
       - name: Release
         uses: softprops/action-gh-release@v2
--- a/.github/workflows/release-awesome-builds.yml
+++ b/.github/workflows/release-awesome-builds.yml
@@ -11,7 +11,7 @@ jobs:
       - name: Merge software/*.yml into software.json via yq
         uses: mikefarah/yq@master
         with:
-          cmd: yq ea '[.] as $item ireduce ([]; . *+ $item )' software/*.yml | yq -o=json -I 0 '{\"softwares\":.}' > software.json
+          cmd: yq ea '[.] as $item ireduce ([]; . *+ $item )' software/*.yml | yq -o=json -I 0 '{"softwares":.}' > software.json
       - name: Raw dump software.json
         run: cat software.json
       - name: Release
--- a/.github/workflows/release-awesome-builds.yml
+++ b/.github/workflows/release-awesome-builds.yml
@@ -1,6 +1,8 @@
 name: Release awesome builds
 
-on: push
+on: 
+  push:
+    branches: [ master ]
 
 jobs:
   build:
@@ -12,7 +14,7 @@ jobs:
         id: get_username
         uses: mikefarah/yq@master
         with:
-          cmd: yq -o=json -I 0 ea '[.] as $item ireduce ([]; . *+ $item )' software/*.yml > software.json
+          cmd: yq ea '[.] as $item ireduce ([]; . *+ $item )' software/*.yml | yq -o=json -I 0 '{\"softwares\":.}' > software.json
       - name: Test
         run: cat software.json
       - name: Release
--- a/.github/workflows/release-awesome-builds.yml
+++ b/.github/workflows/release-awesome-builds.yml
@@ -11,7 +11,7 @@ jobs:
       - name: Merge software/*.yml into software.json via yq
         uses: mikefarah/yq@master
         with:
-          cmd: yq ea '[.] as $item ireduce ([]; . *+ $item )' software/*.yml | yq -o=json -I 0 '{"softwares":.}' > software.json
+          cmd: yq ea '[.] as $item ireduce ([]; . *+ $item )' software/*.yml | yq -o=json -I 2 '{"softwares":.}' > software.json
       - name: Raw dump software.json
         run: cat software.json
       - name: Release
--- a/.github/workflows/release-awesome-builds.yml
+++ b/.github/workflows/release-awesome-builds.yml
@@ -1,8 +1,6 @@
 name: Release JSON for awesome-data builds
 
-on: 
-  push:
-    branches: [ master ]
+on: push
 
 jobs:
   build:
--- a/.github/workflows/release-awesome-builds.yml
+++ b/.github/workflows/release-awesome-builds.yml
@@ -1,0 +1,22 @@
+name: Release awesome builds
+
+on: push
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v5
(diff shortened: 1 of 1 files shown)


Rabenherz112:
I'll throw in my 2 cents; however, I am more than happy to be convinced otherwise. Right now I don't see a strong need to add a consolidated JSON dump to our dataset. Our first-party tooling (e.g. hecat) consumes the YAML directly to produce the already known Markdown/HTML structure, and we don't have a first-party consumer for a "canonical JSON" today.

Even if fully automated via the action, adding another artifact means more to maintain: extra CI steps, docs, and another surface for breakage (I don't expect this ever to break since it's a one-liner, but still).

For users, YAML to JSON is trivial with common tools (such as yq that you used, PyYAML, and others). And in practice, most apps (mine included) don't ingest the raw dataset as-is; they transform, enrich, or filter it to fit their use case (the current PR would also only add software, which would mean it misses all of our categories, license information, and so forth.) So I am not really sure if this dump to JSON will likely add any additional value.

So to me this optimizes for convenience without a demonstrated need. If there's real demand, I'm open to revisiting, but I would like to see a concrete use case and why we should be doing this over the end user / developer. Even using AweXplor as an example, their maintainer didn't acknowledge yet if such a change would help them or if they even wanted / would use it.
Until then, I'd keep the repo clean, without this additional data.

nodiscc:
I agree with the sentiment above:
* brings an extra dependency to non-free github actions (all other CI steps can easlity be replicated using the Makefile, this one cannot)
* limited use (our data is already machine-readable, it is trivial to convert it to JSON using third-party tools)
* doubles the size of each automatic update commit, increases repository size, duplicates information

What would be acceptable is to add a Makefile target `to_json`, people who actually need it could pull the repo and run it manually or from the CI/cron job/... of their choice.

For reference, copy-pasting the raw command does not work for me:

```bash
$ yq ea '[.] as $item ireduce ([]; . *+ $item )' software/*.yml | yq -o=json -I 2 '{"softwares":.}' > software.json
usage: yq [-h] [--yaml-output] [--yaml-roundtrip]
          [--yaml-output-grammar-version {1.1,1.2}] [--width WIDTH]
          [--indentless-lists] [--in-place] [--version]
          [jq_filter] [files ...]
yq: error: argument files: can't open '{"softwares":.}': [Errno 2] No such file or directory: '{"softwares":.}'
usage: yq [-h] [--yaml-output] [--yaml-roundtrip]
          [--yaml-output-grammar-version {1.1,1.2}] [--width WIDTH]
          [--indentless-lists] [--in-place] [--version]
          [jq_filter] [files ...]
yq: error: argument files: can't open '[.] as $item ireduce ([]; . *+ $item )': [Errno 2] No such file or directory: '[.] as $item ireduce ([]; . *+ $item )'
```

jo-chemla:
Thank you both for getting back! 

My fork now daily pulls commits that landed on master branch of upstream `awesome-selfhosted/awesome-selfhosted-data`, and merges them before building the [daily JSON release](https://github.com/jo-chemla/awesome-selfhosted-data/releases) if necessary. The raw JSON is therefore available via this link if anyone want to give it a try https://github.com/jo-chemla/awesome-selfhosted-data/releases/latest/download/software.json - I'll reach out to eg AweXplor to see if this up-to-date JSON is useful. 

Took some time to gather my thoughts, here is some feedback. For the first batch of comments by @Rabenherz112 
 - YAML will always be the source of truth, and the hecat pipelines to build Markdown/HTML from YAML will always be the most useful formats for humans reading the raw data - for humans not using the official or alternative frontends. Regarding machine-readability, the consolidated JSON dump is easier to parse for automated flows than YAML/MD, be it pipelines/scripts or JS logic of frontends built on top of `awesome-selfhosted-data` like those added in PR https://github.com/awesome-selfhosted/awesome-selfhosted-data/pull/1595. There is indeed no first-party consumer fr JSON yet, but this is also probably because the canonical JSON do not exist yet. Demand will probably go in pair with the offering and might at some point become a demonstrated need. 
 - Regarding the JSON builds being an extra CI step, it's true but as you said, this one-liner is pretty self-explanatory, and to me - an outsider of the repo - easier and faster that the existing hecat/makefile. If this additional github action is too much to maintain, I can handle it in my fork to see if strong interest/demand builds around it. It is indeed a one-liner - even simplified compared to initial release, runs in only 1s on the whole dataset. 
 - Regarding the JSON build being not tailored to users/dev, also true, but probably easier to start things off from that JSON rather than the raw YAML or built HTML/MD, and iterate to add additional properties or attributes. `Categories and licenses` can be added to the build pretty easily as well. Edit: Done.
 - It is indeed trivial to convert JSON from YAML via 3rd-party, but this extra-step makes it less easy to consume as-is without the dev/user building this additional translation workflows, while publishing a somewhat official JSON build would facilitate potentially other frontends and any unanticipated uses - like facilitating diff'ing of releases at various dates, to see new additions of the past month etc.

And to the comments by @nodiscc 
 - not sure what you mean by `non-free github actions`. If you want to be able to build this format via a Makefile target `to_json` rather than a github action workflow, this can be done pretty easily. The daily runs of that workflow on my fork are free, and other CI stacks like gitlab also support them. 
 - At the moment these JSONs are published as releases, not commits, with version tag, so no additional duplication of code. Could as well keep a single release, the most up-to-date, and overwrite earlier releases - or keep one per month etc.
 - And the `yq` error you got is probably a case of how bash vs cmd vs powershell handle quotes/double-quotes and escaping. In powershell, you could try `yq ea '[.] as $item ireduce ([]; . *+ $item )' (Get-ChildItem -Name .\sample) | yq -o=json -I 0 '{\"softwares\": .}'`

Hope this helps! Latest workflow visible [here](https://github.com/jo-chemla/awesome-selfhosted-data/blob/master/.github/workflows/release-awesome-builds.yml)

```bash
yq ea '[.] as $item ireduce ([]; . *+ $item ) | {"softwares": .}' software/*.yml -o=json -I=2 > software.json
```


Edit: also published the individual files below, as well as an aggregated https://github.com/jo-chemla/awesome-selfhosted-data/releases/latest/download/awesome-selfhosted-data.json

```json
// awesome-selfhosted-data.json
{
  "softwares": [...],
  "tags": [...],
  "platforms": [...],
  "licenses-nonfree": [...],
  "licenses": [...]
}
```

jo-chemla:
EDIT: I just added the following code block to the github action workflow, which adds the `added_date` flag for all newer awesome-selfhosted-data releases. All softwares which had no date prior to first run of the action will get date of the first run of the github workfow `2025.10.14`. 

Download JSON data for latest release via this link: [awesome-selfhosted-data.json](https://github.com/jo-chemla/awesome-selfhosted-data/releases/latest/download/awesome-selfhosted-data.json) and [awesome-selfhosted-data-extended.json](https://github.com/jo-chemla/awesome-selfhosted-data/releases/latest/download/awesome-selfhosted-data-extended.json). `added_date` is stored in the extended JSON. 

```yaml

      # Extend the JSON awesome-selfhosted-data.json to add eg added_date
      - name: Download previous release JSON
        run: |
          curl https://github.com/jo-chemla/awesome-selfhosted-data/releases/latest/download/awesome-selfhosted-data-extended.json -o latest-release-awesome-selfhosted-data-extended.json
      - name: Merge software/*.yml into software.json via yq
        uses: mikefarah/yq@master
        with:
          cmd: |
            yq eval-all '
              . as $item ireduce ({}; . *+ {
                "softwares": ($item.softwares | map(. + {"added_date": "${{ steps.date.outputs.today }}"}))
              })
            ' latest-release-awesome-selfhosted-data-extended.json software.json > awesome-selfhosted-data-extended.json


```

PR #1612, opened 2025-08-30
update source code URLs

Rabenherz112:
In the latest run of update-metadata i noticed that a lot of repos have changed their source-code.
This PR updates these URLs.

Additionally it re-names PlexRipper to Reaparr (source-code url, name, website, etc. changed)


kokomo123 (approved):
Thank you for the latest updates as always Raben ✅

PR #1624, opened 2025-09-02
Update Drop

### LINKED ISSUE #1623: Drop - "Depends on a proprietary service outside the user's control"
Hi, I'm the maintainer of Drop. I was curious as to why we got flagged with the anti-feature: "Depends on a proprietary service outside the user's control".

The only services that we depend on are metadata services, which are completely optional (users can enter metadata manually, without a service). We even have a couple users running Drop completely offline, within their own LAN network.

quexeky:
Resolves #1623 as Drop does not require any 3rd party services, although they may be added optionally

### DIFF
--- a/software/drop.yml
+++ b/software/drop.yml
@@ -8,7 +8,6 @@ platforms:
   - Docker
 tags:
   - Games - Administrative Utilities & Control Panels
-depends_3rdparty: false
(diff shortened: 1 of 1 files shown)


Rabenherz112:
Waiting for response in #1623 before merge

kokomo123:
Maintainer responded, it has been adjusted appropriately.

PR #1658, opened 2025-09-19
fix target name for strict linting

Rabenherz112:
There was a small bug with make help command.
It returned 2x the argument awesome_lint instead of the correct arguments.


nodiscc (approved):
Thanks

PR #1678, opened 2025-09-24
remove demo from flipt

Rabenherz112:
- ref: #1
- `https://try.flipt.io : HTTPSConnectionPool(host='try.flipt.io', port=443): Max retries exceeded with url: / (Caused by NameResolutionError("<urllib3.connection.HTTPSConnection object at 0x7fb63326d0a0>: Failed to resolve 'try.flipt.io' ([Errno -2] Name or service not known)"))`
- Demo seems to be removed from GitHub and website

### DIFF
--- a/software/flipt.yml
+++ b/software/flipt.yml
@@ -10,7 +10,6 @@ platforms:
   - Go
 tags:
   - Software Development - Feature Toggle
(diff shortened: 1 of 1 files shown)


nodiscc (approved):
Thanks

PR #1688, opened 2025-09-28
Update kuvasz-uptime.yml

adamkobor:


kokomo123 (approved):
Thank you for the fix, appreciate the update description and demo 👍🏻

PR #1711, opened 2025-10-11
docs/pr: change wording for required platforms

Rabenherz112:
Currently, most of the requested changes we do involve the `platform` section. We currently require specifying the main server-side platform, but what's actually relevant is how the software is installed and run, which can (and often does) differ from the underlying server platform.

This PR updates the wording of the checklist item to explicitly clarify this distinction.


kokomo123:
I do think this will clarify a majority of the time where people add the programming languages, and us having to clarify. I'm glad we are doing this now, thank you!

zawgyi112233445566778899-hue:
Good sir ☺️

Issue #1752, opened 2025-10-30
tt-rss is discontinued


Der-Schubi:
https://community.tt-rss.org/t/the-end-of-tt-rss-org/7164

Rabenherz112:
Thanks for the info.
As far as I see, this is more a less a switch of who maintains tt-rss instead of being a total discontinuation.
While the original maintainer Andrew is now gone wn_ and supahgreg seem to have taken over the project. We will need for sure need to update the source code and demo url to reflect that.

I will create PR for it.

Der-Schubi:
You're right, their github seems alive.

PR #1803, opened 2025-11-15
remove The Epube (source unavailable)

Rabenherz112:
- ref: #1
- Repo URL leads to tt-rss github org (https://github.com/tt-rss)
- Website URL is dead `https://tt-rss.org/ZeEpube/ : HTTP 404`
- I can't find the source code in the GtiHub Org either
- Probaly removed with the change mentioned in #1752

Can somebody double-check this?


kokomo123:
Can confirm, seems the epube was discontinued completely and gone. I cannot see the usual links and unfortunately since the main repo is gone, cannot cross reference it with useful-forks, as probably those forks are deleted once the main repo goes kaput.

kokomo123 (approved):
LGTM

nodiscc (approved):

PR #1829, opened 2025-11-28
Update Docker description with additional details

nodiscc:
Expand the description to include information about OCI images and Podman.


kokomo123 (approved):
LGTM, think this will bring more clarity for sure.

PR #1870, opened 2025-12-21
fix yeti-switch demo URL

Rabenherz112:
- ref: #1
- `https://yeti-switch.org/demo.html : HTTP 404`
- A new Demo URL is linked from the main site


kokomo123 (approved):
LGTM, thanks for the fix!

PR #1945, opened 2026-01-17
fix: SUL-1.0 export

Rabenherz112:
In the last PR, I forgot to add the non-free SUL-1.0 license to the hecat configs


kokomo123 (approved):
LGTM, thank you for this.

PR #2098, opened 2026-03-06
Add thealtstack to External Links

kokomo123:
~~even though the original request to add it to the list, this is more of a website that curates it's data openly. I added it to external links as it'd be more appropriate there.~~


Rabenherz112:
The whole project is only a week old, most of these seems to be created via an LLM.
Definitely against adding this.

kokomo123:
Yeah...it was strange how quickly it popped up, so closing, nevermind.

PR #2104, opened 2026-03-07
Update repositories for cannery and memex

shibaobun:
Updating repository locations to Cannery and memEx software


shibaobun:
Thank you!

PR #2146, opened 2026-03-13
tools: update actions/checkout version

Rabenherz112:
- Fixes current deprecation warning we have in our actions by upgrading `actions/checkout` to v6
- Quick check in [actions/checkout](https://github.com/actions/checkout) doesn't seem to show any breaking changes at least
image


ryadulg:
Just

Issue #2209, opened 2026-03-22
CI: WARNING:software_metadata.py: GITLAB_TOKEN environment variable is not set, gitlab projects exist but no metadata will be added/updated


nodiscc:
https://github.com/awesome-selfhosted/awesome-selfhosted-data/pull/2207 was not sufficient

https://github.com/awesome-selfhosted/awesome-selfhosted-data/actions/runs/23410292346/job/68096183394:

```
WARNING:software_metadata.py: GITLAB_TOKEN environment variable is not set, gitlab projects exist but no metadata will be added/updated
```

Even though I added `GITLAB_TOKEN` as an environment variable as described in https://github.com/nodiscc/hecat/pull/152

No metadata have been added in https://github.com/awesome-selfhosted/awesome-selfhosted-data/blob/master/software/jirafeau.yml

Rabenherz112:
Quick check in the docs, you will probably need to define the GITLAB_TOKEN env inside the job
i.e
```yaml
env:
  GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
  GITLAB_TOKEN: ${{secrets.GITLAB_TOKEN}}
```
https://github.com/awesome-selfhosted/awesome-selfhosted-data/blob/master/.github/workflows/daily-update-metadata.yml

https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-secrets#using-secrets-in-a-workflow

nodiscc:
* fix attempt: https://github.com/awesome-selfhosted/awesome-selfhosted-data/commit/ba298a018d897c8d6eebe3e72c529a999c4b5663
* new manual run of the update-metadata workflow https://github.com/awesome-selfhosted/awesome-selfhosted-data/actions/runs/23685926893

Rabenherz112:
https://github.com/awesome-selfhosted/awesome-selfhosted-data/actions/runs/23693077918/job/69023779344
<img width="472" height="397" alt="Image" src="https://github.com/user-attachments/assets/1320d7a1-c601-4ee2-a1d3-00d19dc7f5c5" />
Looks good now

nodiscc:
* https://github.com/awesome-selfhosted/awesome-selfhosted-data/actions/runs/23696086807/job/69031491420

```
INFO:software_metadata.py: processing 27 GitLab projects
INFO:software_metadata.py: processing GitLab batch 1/3 (size: 10)
[...]

PR #2236, opened 2026-03-28
github actions: set environment for update-metadata job

Rabenherz112:
- ref: #2209 

You never actually anywhere defined, that the update-metadata job should use the non-default environment. Therefore the current GitLab token was now empty. Since you said you put the token into production (I can't see it, so that's right, right?), here a second attempt of the fix.


nodiscc:
> Since you said you put the token into production (I can't see it, so that's right, right?)

yes

image

Thanks

PR #2348, opened 2026-04-17
Remove whoogle as EOL

kokomo123:
Per EOL as said by maintainers, it wont receive anymore updates due to Google's cat and mouse game with them. It's a heavy heart to remove this but, at least we can continue to self host it with our own keys. 🫡 

image


nodiscc (approved):

Rabenherz112 (approved):
Awww man

PR #2436, opened 2026-05-08
Add Compass Calendar

tyler-dane:
Adds [Compass Calendar](https://www.compasscalendar.com), a calendar-based task manager.

Thanks for the review =]

A couple notes:
- **demo**: I listed the prod app as the 'demo', because anyone can start using the app immediately without any account signup (data is stored in the browser until they sign up)
- **install guides**: The README summarizes a few ways to get started and also points to full guides in the `docs/` dir in the repo. Setup is also covered on the doc site: https://docs.compasscalendar.com/


tyler-dane:
Hi @Rabenherz112, thanks for the quick review. Can you explain why this was rejected?

PR #2452, opened 2026-05-12
tools(url-check): exclude GitLab repos from URL check

Rabenherz112:
- Currenctly GitLab Repo URLs are still checked via the url-check job
- This is no longer needed, since software_metadata also now supports and performs gitlab checks


kokomo123 (approved):
Fastest approval in the west 💯 🤠 TY for fix

nodiscc (approved):

PR #2593, opened 2026-06-13
remove Bin (unmaintained since May 2025)

Rabenherz112:
- ref: #1
- `ERROR:awesome_lint.py: bin: last updated -384 days, 0:04:43.090728 ago, older than 365 days`
- The software is actually quite minalistic, with only around 300 lines of code, which is why I considerd it for `last_updated_skip`. However, we already have a lot of Bin alternatives, so going for removal for now instead. Other opinions are welcome.


nodiscc (approved):

kokomo123 (approved):

kokomo123:
No issues with this honestly, as much as I love keeping minimalistic projects like this, best for removal. I'm open to revisiting this though to be thrown in ``last_updated_skip`` if we want to 👍🏻

PR #2595, opened 2026-06-13
exclude A Dark Room from unmaintained projects

Rabenherz112:
- ref: #1
- `ERROR:awesome_lint.py: A Dark Room: last updated -386 days, 0:04:43.094674 ago, older than 365 days`
- A Browser web game that is still playable and has no dependencies


nodiscc (approved):
Also an excellent game

PR #2610, opened 2026-06-16
add rss2email to last_updated_skip

nodiscc:
- has been around for more than a decade and works as advertised
- packaged in major linux distributions
- very few dependencies, frequent maintenance not required


Rabenherz112 (approved):
I can testify on the stability as well, as I do run this in some production environments without any issues.

PR #2630, opened 2026-06-23
add Bottle Rocket

gubpalma:

To ensure your Pull Request is dealt with swiftly, please check the following (check the boxes `[x]`):
- [x] Submit one item per pull request. This eases reviewing and speeds up inclusion.
- [x] You have searched the repository for any relevant [issues](https://github.com/awesome-selfhosted/awesome-selfhosted-data/issues) or [PRs](https://github.com/awesome-selfhosted/awesome-selfhosted-data/pulls), including closed ones.
- [x] Any software you are adding is not already listed at any of [awesome-sysadmin](https://github.com/awesome-foss/awesome-sysadmin), [staticgen.com](https://www.staticgen.com/), [staticsitegenerators.bevry.me](https://staticsitegenerators.bevry.me/), [dbdb.io](https://dbdb.io/browse).
- [x] The file you are adding is formatted as described in [addition.md](https://github.com/awesome-selfhosted/awesome-selfhosted-data/blob/master/.github/ISSUE_TEMPLATE/addition.md).
- [x] `Demo` links should only be used for interactive demos, i.e. not video demonstrations. If login credentials are required to access the demo, please link to the credentials directly.
- [x] Comments and unused optional fields have been removed.
- [x] The file you are adding uses [kebab-case](https://en.wikipedia.org/wiki/Letter_case#Kebab_case) file naming, for example `my-awesome-software.yml`.
- [x] Values for `platform` should match the platforms required to install and run the software.
- [x] Any software project you are adding to the list is actively maintained.
- [x] Any software project you are adding was first released more than 4 months ago.
- [x] Any software project you are adding has working installation instructions.
- [x] You understand that your Pull Request will be merged at least ~1 week after approval, depending on maintainers time.


software/bottle-rocket.yml
```diff
@@ -0,0 +1,11 @@
+name: "Bottle Rocket"
+website_url: "https://bottle-rocket.app/"
+description: "Document scanning and management application, compatible with Rocketbook and Scribzee formats"
```
kokomo123:
```suggestion
description: "Document scanning and management application, compatible with Rocketbook and Scribzee formats."
```
Needs a dot at the end here

kokomo123 (dismissed):
As appropriate changes have been made, LGTM pending further thoughts from collaborators. Thank you for your contribution!

software/bottle-rocket.yml
```diff
@@ -0,0 +1,11 @@
+name: "Bottle Rocket"
+website_url: "https://bottle-rocket.app/"
+description: "Document scanning and management application, compatible with Rocketbook and Scribzee formats."
+licenses:
+  - MIT
+platforms:
+  - C#
+  - Docker
+tags:
+  - Note-taking & Editors
```
nodiscc, line 10:
```suggestion
tags:
  - Document Management
```

I'm not sure I understand completely what the program does, it looks like some kind of scanned notes management application https://bottle-rocket.app/docs/docs-general/using-your-server/viewing-your-scans.

@gubpalma could you briefly list the main features of Bottle Rocket?

nodiscc (changes requested):

software/bottle-rocket.yml
gubpalma: Hello there @nodiscc 

This (started out as) an open source replacement for the Rocketbook app, when Rocketbook's future was/is up in the air. 

Rocketbook allows you to write notes on their special books, scan them using the Rocketbook app, and process them in the (closed source) Rocketbook cloud.

Bottle rocket is an effective equivalent, but instead of scanning and sending to the cloud, you send the scans to your own self hosted server. 

A full demo is detailed here
https://bottle-rocket.app/docs/docs-general/tutorial

Apologies for not making this clear

nodiscc: OK

> Rocketbook allows you to write notes on their special books

So it requires this "special book" to work? Can you use plain paper books/make your own compatible books, or do you have to keep purchasing these "proprietary" books from the vendor?

gubpalma: You bought their Rocketbook books with erasable pages, so you no longer need to 'keep' buying them, so to speak.

Recent changes expanded Bottle rocket to be compatible with the Scribzee books as well.

Long term roadmap will likely include more formats.

gubpalma: This is less of a 'document management' application and more focussed to the capture of (hand-written) notes

Rabenherz112: > A [document management system](https://en.wikipedia.org/wiki/Document_management_system) (DMS) is a system used to receive, track, manage and store documents and reduce paper.

Still seems to match then, as you as far as I understand "scan in" these documents to the server. Note-taking & Editors would imply I can actually create notes using the server which as far as I understand is not the case.

gubpalma: @Rabenherz112 initially that's what I thought too, however most of the items under Document Management would be downstream destinations for this app.
i.e. Bottle Rocket scans *images* which are then sent downstream to be managed as documents, or not even treated as documents at all - text from the notes can be sent as emails, etc.

kokomo123 (dismissed):
LGTM with changes made, pending further thoughts from collaborators, thank you for your contribution!

nodiscc:
I'm still unsure about this, see also https://github.com/awesome-selfhosted/awesome-selfhosted-data/issues/2829

gubpalma:
@nodiscc with regards to the scope question, the answer is fairly similar to the linked question - the API is the primary self-hosted component, and everything else is an optional secondary, for convenience.

To clarify:

- The API _can_ be served without an attached web app
- Scans _can_ be sent to the API _without_ using the (accompanying) mobile app

Some simlar models I would cite include Jellyfin, Firefly, and Joplin, all which have optional mobile apps

nodiscc:
The other issue I linked brings up the exact same question: do we want to list projects that are specifically designed to work with specific, closed hardware products.

kokomo123:
I don't think we should personally open up the door for specific, hardware projects. it means shutting out potential new projects for their niche devices. but i think we need to stay true to our roots of this list. im open to other opinions

Rabenherz112:
I thought about this now over an hour. And I am prob. misreading the whole thing, but I don't think grouping this in with #2829 is the right way.

Now first of all I think this and #2829 are two completely different case in my opinion. Bottle Rocket is normal server software which you deploy on whatever Docker host you like and Rocketbook/Scribzee are just supported input formats, not something the hosting depends on while STR only runs on the speaker, so you genuinely can't self-host it "wherever you want." That's much closer in spirit to our "depends on a specific cloud provider" exclusion, just with a box instead of an PaaS/API. (Also I don't think anybody is browsing this list to find a fix for their SoundTouch speaker anyway, so I don't think this adds any value (more of a "wrong place for it" problem))

> I don't think we should personally open up the door for specific, hardware projects

On the general question, I'd rather we did not add a "specific closed hardware" exclusion.

The way I see this list, it exists to show alternatives to proprietary software and cloud services that you pay for with your data (if not with money), and ways to keep actually using a product (or extend the use-case) you own the way you want to rather than the way the vendor decided.
We also already list software for proprietary hardware and I think those are a good fit in this list. Ownfoil for organizing Nintedo Switch Games (and I think Nitendo is one of the best example of closed hardware products), TeslaMate to export their own car's data and if I prob. look a bit more I would have found quite a bit more. Also a vendor shutting down the software/service behind hardware you already paid for is probably the strongest argument there is for self-hosting something.

That said listing stuff like that should still come to a sensible conclusion. I don't think we should list every project that solves an Issue for one speific Hardware (nobody is gonna go to this list anyway to find it), if it is not a very popular thing. So maybe the questions is not "does it depend on hardware," but "is the hardware common enough to be worth listing generally."

Though I am open to not listing stuff like this and not having to deal with the headache if the majority thinks so. Just wanting to put out at least my thoughts to this.

kokomo123:
After reading the thoughts above, I am not sure how best to proceed. While we do have those specific projects sure it makes sense but yeah, it's having the sense of this generally. 

> That said listing stuff like that should still come to a sensible conclusion. I don't think we should list every project that solves an Issue for one speific Hardware (nobody is gonna go to this list anyway to find it), if it is not a very popular thing. So maybe the questions is not "does it depend on hardware," but "is the hardware common enough to be worth listing generally."

Tesla and Nintendo switches are common hardware and are in abundance. It's just having loads of niche use cases that pile on the list that i'm worried about, I am happy to with this now fully understanding what bottle rocket is. but i'll wait before approving

PR #2726, opened 2026-07-14
Add Foliofox

unav4ila8le:
Foliofox is a self-hosted net worth and portfolio tracker with AI-powered financial insights, scenario planning, time-weighted returns, multi-currency support, and broker import. It is actively maintained and has been publicly available for well over 4 months.


unav4ila8le:
Hi @nodiscc , thanks for looking at the PR. May I know why it wasn't merged?

PR #2878, opened 2026-08-10
add Flutter, update AppFlowy

lukepighetti:
adding Flutter and updating AppFlowy to be more specific. (will be contributing more awesome selfhosted Flutter projects in the near future)


nodiscc:
How are flutter applications distributed? Are they standalone binaries? Do they need a specific runtime/middleware?

`platforms` are the server-side requirements to run the software. Does running programs written with Flutter imply installing a specific piece of software, or are they self-sufficient?

lukepighetti:
hey thanks so much for the review and the question. flutter would be the frontend component attached to a backend service. so in the case of appflowy the server component is written in Rust but in order to make use of it the Flutter frontend is required. it completes the self-hosted product. 

in some cases the Flutter web app is baked into the binary. in some cases something like caddy is serving it as part of a docker-compose file. and in some cases the Flutter frontend binary is made available to be installed on a desktop or mobile device

hope that is helpful and please let me know if there are any other questions

nodiscc:
In this case, there are no specific-server side requirements. As long as the framework/language/... compiles to standard HTML/CSS/JS it is not necessary or desired to list it in the `platforms` list.

`platforms` are the server-side requirements to run the software.

I think we should close this.

Maybe `Dart` should be removed as well?

lukepighetti:
you make a good point about flutter, i will close this and fix my other PRs. Dart should stay as a platform IMO because there are self-hosted apps that use Dart for the backend service

Issue #1699, opened 2025-10-07
Add RUSTDESK


demo308:
Thanks for taking the time to suggest an addition to awesome-selfhosted! 

Please fill out information below (all fields are mandatory unless noted otherwise):

```yaml
# software name
name: "Rustdesk"
# URL of the software project's homepage
website_url: "https://rustdesk.com"
# URL where the full source code of the program can be downloaded
source_code_url: "https://github.com/rustdesk/rustdesk"
# description of what the software does, shorter than 250 characters, sentence case
description: "Remote Desktop."
# list of license identifiers, see https://github.com/awesome-selfhosted/awesome-selfhosted-data/blob/master/licenses.yml for the full list of licenses
licenses:
  - Apache-2.0
  - AGPL-3.0
# list of languages/platforms, see https://github.com/awesome-selfhosted/awesome-selfhosted-data/tree/master/platforms for the full list of platforms
platforms:
  - Java
  - Python
  - PHP
  - Nodejs
  - Deb
  - Docker
# list of tags (categories), see https://github.com/awesome-selfhosted/awesome-selfhosted-data/tree/master/tags for the full list of tags
tags:
  - Remote Desktop
# (optional, true/false, default false) whether the software depends on a third-party service outside the user's control
depends_3rdparty: true
# (optional) link to an interactive demo of the software
demo_url: "https://my.awesome.softwar.e/demo"
# (optional) link to a list of clients/addons/plugins/apps/bots... for the software
related_software_url: "https://my.awesome.softwar.e/apps"
```

To ensure your issue is dealt with swiftly, please check the following (check the boxes `[x]`):
- [ ] Submit one item per issue. This eases reviewing and speeds up inclusion.
- [ ] You have searched the repository for any relevant [issues](https://github.com/awesome-selfhosted/awesome-selfhosted-data/issues) or [PRs](https://github.com/awesome-selfhosted/awesome-selfhosted-data/pulls), including closed ones.
- [ ] Any software you are adding is not already listed at any of [awesome-sysadmin](https://github.com/awesome-foss/awesome-sysadmin), [staticgen.com](https://www.staticgen.com/), [staticsitegenerators.bevry.me](https://staticsitegenerators.bevry.me/), [dbdb.io](https://dbdb.io/browse).
- [ ] Any software project you are adding to the list is actively maintained.
- [ ] Any software project you are adding was first released more than 4 months ago.
- [ ] Any software project you are adding has working installation instructions.

[...]

PR #2879, opened 2026-08-10
add RustDesk (on #2878)

lukepighetti:
this is meant to merge after https://github.com/awesome-selfhosted/awesome-selfhosted-data/pull/2878

NOTE: RustDesk is now a good fit for awesome-selfhosted because it is now [self-hosted](https://github.com/awesome-selfhosted/awesome-selfhosted-data/issues/1699#issuecomment-3618002166)


Technetium1 (changes requested):
`[4]` in flutter description should be removed. `rust-desk.yml` should be renamed `rustdesk.yml`

lukepighetti:
> `[4]` in flutter description should be removed. `rust-desk.yml` should be renamed `rustdesk.yml`

done, thank you for the review!

lukepighetti:
@Technetium1 got this cleaned up based on comments in #2878. thank you!

Technetium1 (approved):

Rabenherz112:
So first of all additions should use the PR Template

>  If you are adding new software to the list, DO NOT DELETE THE TEXT BELOW . Please make sure relevant boxes are checked [x] 

Also RustDesk was listed here, but removed due to

> The problem isn't the server part (no matter if on docker or not) but that you can't point the client to your server. Unless you either pay for the client generator or by manually copy pasting it into the client. Which is not really feasible for what the product wants to be, a remote support tool.

See https://github.com/awesome-selfhosted/awesome-selfhosted-data/pull/1404

Was this Issue solved?

nodiscc:
> Unless you either pay for the client generator or by manually copy pasting it into the client

This is still true, the default branded client does connect to admin.rustdesk.com and api.rustdesk.com as defaults. It can be disabled manually. You can still manually configure the client to connect to your own server, or you can build your own custom-branded client with a modified default server.

Personally I will not approve its addition. A remote desktop client is too sensitive to consider a third-party relay as the default, hidden in the settings, as acceptable. Opt-in would probably change my view. SSO is also gated behind the Pro version offering. 

Surely there must be better alternatives, but I haven't looked.

Rabenherz112:
> Personally I will not approve its addition. A remote desktop client is too sensitive to consider a third-party relay as the default, hidden in the settings. 

Agreed.

kokomo123:
> > Unless you either pay for the client generator or by manually copy pasting it into the client
> 
> This is still true, the default branded client does connect to admin.rustdesk.com and api.rustdesk.com as defaults. It can be disabled manually. You can still manually configure the client to connect to your own server, or you can build your own custom-branded client with a modified default server.
> 
> Personally I will not approve its addition. A remote desktop client is too sensitive to consider a third-party relay as the default, hidden in the settings, as acceptable. Opt-in would probably change my view. SSO is also gated behind the Pro version offering.
> 
> Surely there must be better alternatives, but I haven't looked.

I agree with these reasons wholeheartedly, i do not in good conscious recommend this project to awesome-selfhosted.

PR #2905, opened 2026-08-14
add Awesome Tunneling to external links in VPN tag header

nodiscc:
This is a well maintained project which lists alternatives to ngrok (self-service reverse proxy, allows clients to expose their locally running services over a public subdomain). I used it recently (ended up setting up `pgrok` - example `pgrok http 8080` exposes the service running a laptop on 127.0.0.1:8080 as `https://mysuername.pgrok.mydomain.org`)

I havent added it to the `Any software you are adding is not already listed at any of...` list in the PR template/contribution guidelines because there is probably some overlap between both lists (it goes a bit into VPN territory).

However it is better suited to submissions such as this one https://github.com/awesome-selfhosted/awesome-selfhosted-data/pull/2901

### DIFF
--- a/tags/vpn.yml
+++ b/tags/vpn.yml
@@ -3,3 +3,6 @@ description: A [virtual private network (VPN)](https://en.wikipedia.org/wiki/Vir
 redirect:
   - title: awesome-sysadmin/VPN
     url: https://github.com/awesome-foss/awesome-sysadmin#vpn
+external_links:
+  - title: Awesome-Tunneling
+    url: https://github.com/anderspitman/awesome-tunneling


Rabenherz112 (commented):
I am fine with adding the list itself, though I'm not sure the external links section is the right place for it. Currently this section is used only for resources that are not catalogs of self-hostable software. Otherwise we use the `redirect` to point to another awesome-list.

Would't it make more sense to create a new tag Tunneling, where we redirect to this list?

kokomo123:
that's a interesting point here raben. Thus the tag creation keeps expanding 👍🏻

kokomo123:
I'll approve this PR once the appropriate tag has been made.

nodiscc:
I've thought a bit about this, I'm not sure a `Tunnelling` category/tag would be relevant or maintainable. Tunneling encompasses much more than this (VPNs and SSH can provide tunneling-related functionality, yet still we _do_ list them here).

The wikipedia page for ngrok (the OG proprietary service that provides this _specific_ feature set) describes it as:

> software platform that provides secure tunneling services, enabling locally hosted web applications and services to be exposed to the public internet through temporary or static public URLs

These solutions are a basically combination of reverse proxy and VPN/SSH-like tunnels, bolting administration tools and authentication on top.

We could consider them as as specific type of reverse proxy and accept them in this section, but then most programs already listed on awesome-tunneling would qualify, and we would just duplicate a large part of their work (what we've been doing from the start, if a better resource already exists, is to delegate to it)

My proposal is specifically about forwarding such additions to awesome-tunneling, manually (there are not many), and add it to "external_links" links where needed (in the VPN tag would be a good start). You're right, the global `External Links` is not the correct place to add this, I will change it.

Rabenherz112 (approved):
Should prob. also added as a external link in awesome-sysadmin VPN Section

PR #3044, opened 2026-09-06
tools(url-check): exclude always failing urls

Rabenherz112:
This adds an additional badge of urls which always fail due to some DDos / Bot Protection / something else blocking the check tool itself and should be excluded from the checks


kokomo123 (approved):
Thank you for the fix!

PR #3045, opened 2026-09-06
remove demo of openreader

Rabenherz112:
OpenReader has geo-restricted their main webpage and demo to USA access only.
Update all references and links to point exclusively to the globally accessible documentation page instead of the geo-restricted page.


kokomo123 (approved):
LGTM, thank you for the fix!

PR #3046, opened 2026-09-06
update descriptions to remove 'open-source'

Rabenherz112:
Removes any mentions of 'open-source' from all descriptions. Being open-source is already implied when listed here, and in some cases the the software wasn't even (libre-) free.


kokomo123 (approved):
LGTM, thank you for the fix!

